PR #21, opened 2014-11-25
(PDOC-17) Add support for YARD tags in puppet code

hkenney:
Prior to this commit, most of the HTML created for documenting
puppet code was auto-generated by YARD, meaning that it was not
always presented in a way that was consistent or easy to read. 

Now we are explicitly defining the HTML and styling that we want
to use when generating documentation for defined types and classes.
This means that we can define exactly how we want the output to look
so that is more compatible with puppet code.

I have yet to add links to any of the details around parameters because it's
not clear to me where those links should go at the moment. It should be
simple to add them though when we decide what we want them to look like.

### DIFF
--- a/lib/puppetx/puppetlabs/strings/yard/templates/default/definedtype/setup.rb
+++ b/lib/puppetx/puppetlabs/strings/yard/templates/default/definedtype/setup.rb
@@ -17,6 +17,18 @@ def parameter_details
   erb(:parameter_details)
 end
 
+def header
+  if object.type == :hostclass
+    @header_text = "Puppet Class: #{object.name}"
+  elsif object.type == :definedtype
+    @header_text = "Puppet Defined Type: #{object.name}"
+  else
+    @header_text = "#{object.name}"
+  end
+
+  erb(:header)
+end
+
 def docstring
   examples = Hash.new
   example_tags = object.tags.find_all { |tag| tag.tag_name == "example" }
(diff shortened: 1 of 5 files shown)


lib/puppetx/puppetlabs/strings/yard/templates/default/definedtype/setup.rb
```diff
@@ -1,10 +1,92 @@
...
+#
+# @param params_hash [Array] parameter details obtained programmatically
+# @param tags_hash [Array] parameter details obtained from comments
+#
+# @return [Hash] The relevant information about each parameter
+# @option opts [String] :name The name of the parameter
+# @option opts [String] :fq_name The fully qualified parameter name
+# @option opts [String] :desc The description provided in the comment
+# @options opts [Array] :types The parameter type(s) specified in the comment
+# @options opts [Boolean] :exists? True only if the parameter actually exists and just not just defined in the comment
```
hlindberg: "actually" and "just not just", suggest writing something like "True only if the parameter exists in the documented logic and not just in a comment"

lib/puppetx/puppetlabs/strings/yard/templates/default/definedtype/setup.rb
```diff
@@ -1,10 +1,92 @@
...
+    param_types = param_tag.nil? ? nil : param_tag.types
+
+    parameter_info.push({:name => param[0], :fq_name => param[1], :desc => description, :types => param_types, :exists? => true})
+  end
+
+  # Check if there were any comments for parameters that do not exist
+  tags_hash.each do |tag|
+    param_exists = false
+    parameter_info.each do |parameter|
+      if parameter.has_value?(tag.name)
```
hlindberg: This looks a bit odd - if I am reading this right, it means that a parameter is considered to exist if one of the values for :name, :fq_name, :desc, :types, or :exists? have the same value as tag.name. This could potentially go wrong (??) if the description text contains a matching name (?). Was this done to match both :name and :fq_name?

PR #22, opened 2015-01-23
(PDOC-24) Add basic templates for functions

hkenney:
Prior to this commit, all the HTML for puppet functions was just
generated by YARD as if it were a regular method. Since we ultimately
want to have our own custom templates, lay out the groundwork for customizing
the templates for functions and update the page headers to be more
readable.


lib/puppetx/puppetlabs/strings/yard/templates/default/html_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+        result << "<strike>"
+      end
+
+      result << "<span class=\"name\">#{param[:name]} </span>"
+      result << "<span class=\"type\">"
+
+      if param[:types]
+        result << "(" << "<tt>" << param[:types].join(", ") << "</tt>" << ")"
+      else
+        result << "(<tt>TBD</tt>)"
```
hlindberg: what is this TBD ?

lib/puppetx/puppetlabs/strings/yard/templates/default/html_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+        result << "</span>"
+
+      # This is only relevant for manifests, not puppet functions
```
hlindberg: it is worth a brief explanation why it is not relevant

lib/puppetx/puppetlabs/strings/yard/templates/default/html_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  # Generates the HTML to format the relevant data about parameters
+  def generate_parameters(params)
+    result = []
+
+    params.each do |param|
+      result << "<li>"
+
+      if !param[:exists?]
```
hlindberg: it is worth commenting which parameters that are given <strike> style

lib/puppetx/puppetlabs/strings/yard/templates/default/html_helper.rb
hkenney: When we discussed how to deal with parameters we decided that if the user did not specify a type we would list it as "TBD" for puppet manifests. I decided to keep it consistent for functions as well, but I'm happy to change it if you think something else would be more appropriate

lib/puppetx/puppetlabs/strings/yard/templates/default/puppetnamespace/setup.rb
```diff
@@ -1,7 +1,87 @@
...
+  @methods = object.children
+  @template_helper = TemplateHelper.new
+end
+
+def header
+  if @methods[0]['puppet_4x_function']
+    @header_text = "Puppet 4 Functions"
+  else
+    @header_text = "Puppet 3 Functions"
+  end
```
hlindberg: What happens above if there are both 3x functions and 4x functions ? (unsure how 'header' is used and what is in @methods, more than one (it is an array after all) ?

lib/puppetx/puppetlabs/strings/yard/templates/default/puppetnamespace/setup.rb
```diff
@@ -1,7 +1,87 @@
...
+
+  erb(:header)
+end
+
+def box_info
+  @source_files = []
+
+  @methods.each do |method|
+   # extract the file name and line number for each method
```
hlindberg: alignment off by one char

lib/puppetx/puppetlabs/strings/yard/templates/default/puppetnamespace/setup.rb
hkenney: So how I _think_ this works is that this template will be used twice, once to generate dos for all the puppet 3 functions (which will all be displayed on one webpage) and once to generate all the docs for the puppet 4 functions (which will all be displayed on another webpage). When I poked around in the debugger it seemed that this template was used twice, and each time the array contained only 3x or 4x functions.

This is 100% just from observation though so I could be wrong

lib/puppetx/puppetlabs/strings/yard/templates/default/template_helper.rb
```diff
@@ -0,0 +1,81 @@
...
+  # @return [Hash] The relevant information about each parameter
+  # @option opts [String] :name The name of the parameter
+  # @option opts [String] :fq_name The fully qualified parameter name
+  # @option opts [String] :desc The description provided in the comment
+  # @options opts [Array] :types The parameter type(s) specified in the comment
+  # @options opts [Boolean] :exists? True only if the parameter exists in the documented logic and not just in a comment
+  def extract_param_details(parameters, tags_hash, fq_name = false)
+    parameter_info = []
+
+
```
hlindberg: extra blank line

lib/puppetx/puppetlabs/strings/yard/templates/default/template_helper.rb
```diff
@@ -0,0 +1,81 @@
...
+    # Extract the information for parameters that actually exist
```
hlindberg: as opposed to which parameters? (I think I know the answer, but it would be nice if the comment gave more explanation to the next engineer)

lib/puppetx/puppetlabs/strings/yard/templates/default/template_helper.rb
```diff
@@ -0,0 +1,81 @@
...
+  # useful information and returns it as an array of hashes which can
+  # be printed and formatted as HTML
+  #
+  # @param parameters [Array] parameter details obtained programmatically
+  # @param tags_hash [Array] parameter details obtained from comments
+  # @param fq_name [Boolean] does this paramter have a fully qualified name?
+  #
+  # @return [Hash] The relevant information about each parameter
+  # @option opts [String] :name The name of the parameter
+  # @option opts [String] :fq_name The fully qualified parameter name
```
hlindberg: fq_name is documented to be a String, but has a Boolean default value - confusing

lib/puppetx/puppetlabs/strings/yard/templates/default/template_helper.rb
```diff
@@ -0,0 +1,81 @@
...
+  # @option opts [String] :desc The description provided in the comment
+  # @options opts [Array] :types The parameter type(s) specified in the comment
+  # @options opts [Boolean] :exists? True only if the parameter exists in the documented logic and not just in a comment
```
hlindberg: the documentation refers to `opts` but there is no such parameter

lib/puppetx/puppetlabs/strings/yard/templates/default/html_helper.rb
hlindberg: 3x functions will never have type info per parameter, they only describe the arity. So for those it will be very irritating to see TBD. We cannot really say anything - if we use `Any` it may be wrong (the function may only accept something else - there is no way to know). So for 3x functions I think we should leave this blank. In other cases 'TBD' is good enough for now - someone may want to change it later to 'unspecified' or something like that. We could run this by UX later, and ask for feedback.

lib/puppetx/puppetlabs/strings/yard/templates/default/puppetnamespace/setup.rb
hlindberg: ok, maybe just add a comment that says `# The list is expected to only contain one type of functions` to help the next reader understand that the logic is not flawed.

hkenney:
Okee dokee, I shall fix the issue around the @param tag for 3x functions and update the comments! :+1:

lib/puppetx/puppetlabs/strings/yard/templates/default/template_helper.rb
hkenney: The `opts` is how YARD suggests that you document a Hash (http://www.rubydoc.info/gems/yard/file/docs/Tags.md#option). The mix of @option and @options is just a typo, I'll fix that.

hlindberg: yeah, for a parameter, but you are using options to describe the returned hash - do not think that works. Each @option has a reference to the @param it describes a detail of. There is no way I can see to make it refer to the returned value.

hkenney: Ah I see what you mean, do you have a suggestion for how to properly document a hash if it's a return value?

hlindberg: there is none in yard at least - you have to write it in the @return tag i.e. "A hash with the following keys and value types `{:name => String, :... }`
(best I can come up with - sorry)

lib/puppetx/puppetlabs/strings/yard/templates/default/template_helper.rb
```diff
@@ -0,0 +1,107 @@
...
+  # Given the parameter information and YARD param tags, extracts the
+  # useful information and returns it as an array of hashes which can
+  # be printed and formatted as HTML
+  #
+  # @param parameters [Array] parameter details obtained programmatically
+  # @param tags_hash [Array] parameter details obtained from comments
+  # @param fq_name [Boolean] does this parameter have a fully qualified name?
+  #
+  # @return [Hash] The relevant information about each parameter with the following keys/values:
+  # {:name    => [String] The name of the parameter
```
hlindberg: think you need to indent the continuation of the @return to make it appear in the output for it

lib/puppetx/puppetlabs/strings/yard/templates/default/template_helper.rb
```diff
@@ -0,0 +1,107 @@
...
+  # Generates parameter information in situations where the information can only
+  # come from YARD tags in the comments, not from the code itself. For now the only
+  # use for this is 3x functions. In this case exists? will always be true since we
+  # cannot verify if the parameter exists in the code itself. We must trust the user
+  # to provide information in the comments that is accurate.
+  #
+  # @param param_tags [Array] parameter details obtained from comments
+  #
+  # @return [Hash] The relevant information about each parameter with the following keys/values:
+  # {:name          => [String] The name of the parameter
```
hlindberg: same thing here

hlindberg:
Looks fine, will merge when travis goes green.

PR #24, opened 2015-02-04
[DO NOT MERGE] (PDOC-26) Rename Puppetx to PuppetX

hkenney:
Due to the changes in PUP-3900 which renamed Puppetx to PuppetX
and puppetx to puppet_x, strings was failing since the namespace
had not been updated in strings. In order to be compatible with this
change, update strings by renaming the namespace to match.

This will prevent strings from working with anything older than Puppet 4
however and will require a version bump.

[DO NOT MERGE]
This will break strings will all versions of puppet older than 4.0,  so
we want to wait on merging this change until Puppet 4 is released and
we bump the strings version.


hkenney:
We've adjusted our strategy for dealing with this a bit so I'm going to go ahead and close this pull request. Will open a new one shortly.

PR #26, opened 2015-02-12
(PDOC-27) Don't require options for 3x functions

hkenney:
Prior to this commit, the 3x function handler assumed that at
least two arguments (the name and one or more additional arguments)
were passed into newfunction when creating a 3x function. However
that is not actually required, the only argument needed is the name.

Update the 3x function handler so that it will not throw and exception
if only one argument is given.


hkenney:
I fixed the rspec version issue, but now it looks like I now have legitimate failures for Ruby 2x. ಠ_ಠ Looking into these now

PR #27, opened 2015-03-12
(PDOC-25) Fix mangled puppet namespaces

hkenney:
Prior to this commit, puppet namespaces would be totally mangled when
processed by YARD since YARD is looking for Ruby namespaces where the
first letter of each segment must be uppercase.

In order to fix this, patch the relevant regular expression constants
in YARD so that they will accept segments that begin with uppercase
and lowercase letters.

An upstream pull request has been accepted into YARD to add another
regex constant which is necessary to complete this patch. However until
the next version of the YARD gem is released, we must include a temporary
hack to fix the problematic method in question (as it is not yet using
the added variable). Once the new version is released, the conditional
statement in code_ext/yard.rb can be removed entirely and all we will
need to do is patch the two regex constants.

### DIFF
--- a/lib/puppet_x/puppetlabs/strings/yard/core_ext/yard.rb
+++ b/lib/puppet_x/puppetlabs/strings/yard/core_ext/yard.rb
@@ -1,0 +1,40 @@
+require 'yard'
+
+require 'puppet_x/puppetlabs/strings'
+
+# Patch the regular expression used to match namespaces
+# so it will allow namespace segments that begin with
+# both uppercase and lowercase letters (i.e. both
+# Puppet::Namespace and puppet::namespace)
+YARD::CodeObjects.send(:remove_const, :CONSTANTMATCH)
+YARD::CodeObjects::CONSTANTMATCH = /[a-zA-Z]\w*/
+
+# This is a temporary hack until a new version of YARD is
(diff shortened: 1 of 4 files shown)


hkenney:
Actually, please don't merge this just yet. I want to add a quick test to prevent regressions.

hkenney:
Okay, this is ready for review! :+1:

lib/puppet_x/puppetlabs/strings/yard/core_ext/yard.rb
```diff
@@ -0,0 +1,40 @@
+require 'yard'
+
+require 'puppet_x/puppetlabs/strings'
+
+# Patch the regular expression used to match namespaces
+# so it will allow namespace segments that begin with
+# both uppercase and lowercase letters (i.e. both
+# Puppet::Namespace and puppet::namespace)
+YARD::CodeObjects.send(:remove_const, :CONSTANTMATCH)
+YARD::CodeObjects::CONSTANTMATCH = /[a-zA-Z]\w*/
```
hlindberg: I think these two lines should not be here. It means always setting the constant. A couple of lines down, you will then remove it, and set it again and the patch does not come into effect. (Or so it seems).

PR #30, opened 2015-06-23
Use different debuggers for ruby 2 and 1.9

iankronquist:
The debugger gem does not support ruby 2.0.
The byebug gem does not support ruby < 2.0.
Conditionally specify the right gem for the right version of ruby.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,5 +26,9 @@ end
 
 group :development do
   gem 'pry'
-  gem 'pry-debugger'
+  if RUBY_VERSION[0..2] == '1.9'
+    gem 'pry-debugger'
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -26,5 +26,9 @@ end
 
 group :development do
   gem 'pry'
-  gem 'pry-debugger'
+  if RUBY_VERSION[0..2] == '1.9'
+    gem 'pry-debugger'
+  elsif RUBY_VERSION[0] == '2'
+    gem 'pry-byebug'
+  end
```
ferventcoder: Looks similar to what I do.

ferventcoder: :+1:

PR #31, opened 2015-06-24
(PDOC-37) Warn when documented name does not match declared name

iankronquist:
A rude first attempt. Let's see how this goes.

For warnings we just print to stderr. Note that Yard does not print its warnings to stderr, which is a bit crude.

Note that this does not do any type checking. Presently, if I understand correctly, our handlers do not add any type information so we can't compare types with the comments.
Addresses [PDOC-37](https://tickets.puppetlabs.com/browse/PDOC-37)

### DIFF
--- a/lib/puppet/face/strings.rb
+++ b/lib/puppet/face/strings.rb
@@ -50,6 +50,7 @@ Puppet::Face.define(:strings, '0.0.1') do
       # For now, assume the remaining positional args are a list of manifest
       # and ruby files to parse.
       yard_args = (args.empty? ? MODULE_SOURCEFILES : args)
+      class YARD::Logger; def progress(*args); end; end
 
       yardoc_actions.generate_documentation(*yard_args)
     end
--- a/lib/puppet/face/strings.rb
+++ b/lib/puppet/face/strings.rb
@@ -50,7 +50,11 @@ Puppet::Face.define(:strings, '0.0.1') do
       # For now, assume the remaining positional args are a list of manifest
       # and ruby files to parse.
       yard_args = (args.empty? ? MODULE_SOURCEFILES : args)
-      class YARD::Logger; def progress(*args); end; end
+
+      # This line monkeypatches yard's progress indicator so it doesn't write
+      # all over the terminal. This should definitely not be in real code, but
+      # it's very handy for debugging with pry
+      #class YARD::Logger; def progress(*args); end; end
 
       yardoc_actions.generate_documentation(*yard_args)
     end
--- a/lib/puppet_x/puppetlabs/strings/yard/templates/default/template_helper.rb
+++ b/lib/puppet_x/puppetlabs/strings/yard/templates/default/template_helper.rb
@@ -36,7 +36,7 @@ class TemplateHelper
   #    :desc    => [String] The description provided in the comment
   #    :types   => [Array] The parameter type(s) specified in the comment
   #    :exists  => [Boolean] True only if the parameter exists in the documented logic and not just in a comment}
-  def extract_param_details(parameters, tags_hash, locations, fq_name = false)
+  def extract_param_details(parameters, tags_hash, fq_name = false)
     parameter_info = []
 
     # Extract the information for parameters that exist
@@ -73,13 +73,6 @@ class TemplateHelper
       end
       if !param_exists
         parameter_info.push({:name => tag.name, :desc => tag.text, :types => tag.types, :exists? => false})
-        if locations.length >= 1 and locations[0].length == 2
-          file_name = locations[0][0]
-          line_number = locations[0][1]
-          $stderr.puts "[warn]: The parameter #{tag.name} is documented, but doesn't exist in your code, in file #{file_name} near line #{line_number}"
-        else
-          $stderr.puts "[warn]: The parameter #{tag.name} is documented, but doesn't exist in your code. Sorry, the file and line number could not be determined."
-        end
       end
     end
 
@@ -110,5 +103,27 @@ class TemplateHelper
     end
 
     parameter_info
+  end
+
+  # Check that the actual function parameters match what is stated in the docs.
+  # If there is a mismatch, print a warning to stderr.
+  # This is necessary for puppet classes and defined types. This type of
+  # warning will be issued for ruby code by the ruby docstring parser.
+  # @param object the code object to examine for parameters names
+  def check_parameters_match_docs(object)
+    param_tags = object.tags.find_all{ |tag| tag.tag_name == "param"}
+    names = object.parameters.map {|l| l.first.gsub(/\W/, '') }
+    locations = object.files
+    param_tags.each do |tag|
+      if not names.include?(tag.name)
+        if locations.length >= 1 and locations[0].length == 2
+          file_name = locations[0][0]
+          line_number = locations[0][1]
+          $stderr.puts "[warn]: The parameter #{tag.name} is documented, but doesn't exist in your code, in file #{file_name} near line #{line_number}"
+        else
+          $stderr.puts "[warn]: The parameter #{tag.name} is documented, but doesn't exist in your code. Sorry, the file and line number could not be determined."
+        end
+      end
+    end
   end
 end
--- a/lib/puppet_x/puppetlabs/strings/yard/templates/default/template_helper.rb
+++ b/lib/puppet_x/puppetlabs/strings/yard/templates/default/template_helper.rb
@@ -1,7 +1,8 @@
+require "puppet"
+
 # A class containing helper methods to aid in the extraction of relevant data
 # from comments and YARD tags
 class TemplateHelper
-
   # Extracts data from comments which include the supported YARD tags
   def extract_tag_data(object)
     examples = Hash.new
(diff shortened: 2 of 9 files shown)


lib/puppet_x/puppetlabs/strings/yard/templates/default/template_helper.rb
```diff
@@ -73,9 +75,17 @@ def extract_param_details(parameters, tags_hash, fq_name = false)
...
+          line_number = location[0][1]
+          $stderr.puts "[warning] #{file_name}: #{line_number} The parameter #{tag.name} is documented, but doesn't actually exist in your code."
+        else
+          $stderr.puts "[warning] Unknown file name or line number: The parameter #{tag.name} is documented, but doesn't actually exist in your code."
+        end
       end
     end
-
+    #puts "params #{parameters} tags_hash #{tags_hash}"
+    #require 'pry'; binding.pry
```
hkenney: Don't forget to remove your commented out debug statements ;)

lib/puppet_x/puppetlabs/strings/yard/templates/default/template_helper.rb
```diff
@@ -73,9 +75,17 @@ def extract_param_details(parameters, tags_hash, fq_name = false)
       end
       if !param_exists
         parameter_info.push({:name => tag.name, :desc => tag.text, :types => tag.types, :exists? => false})
+        if location.length < 1 and location[0].length == 2
+          file_name = location[0][0]
+          line_number = location[0][1]
+          $stderr.puts "[warning] #{file_name}: #{line_number} The parameter #{tag.name} is documented, but doesn't actually exist in your code."
```
hlindberg: what is the difference between "actually exist" and just "exist"? Take out the word "actually"

lib/puppet_x/puppetlabs/strings/yard/templates/default/template_helper.rb
```diff
@@ -73,9 +75,17 @@ def extract_param_details(parameters, tags_hash, fq_name = false)
...
+        else
+          $stderr.puts "[warning] Unknown file name or line number: The parameter #{tag.name} is documented, but doesn't actually exist in your code."
```
hlindberg: The same thing here (actually :-) )

lib/puppet_x/puppetlabs/strings/yard/templates/default/template_helper.rb
iankronquist: Oops, sorry. Fixed.

iankronquist:
What are peoples thoughts on disabling the Rubocop errors ` Metrics/AbcSize` and `Metrics/PerceivedComplexity`? I'm all for linters making my code pretty, but these are just rubocop guessing at how complex the code is. I think it's mainly grouchy about all those ternary operators.

In any case, this PR does not introduce new linting errors.

iankronquist:
So apparently rubocop doesn't check templates. I called it on this file explicitly. Never mind.

hkenney:
@iankronquist I think we should squash the most recent three commits (removing debug statements, updating warning messages, and fixing rubocop errors). 

You can either squash them into your first commit, or sometimes what we do is have a commit for addressing pull request feedback where you can just describe all the things you changed.

But ultimately nothing in those commits is very interesting on its own, so it will make the git history a little easier for people to navigate in the future if we squash those.

hlindberg:
We should have the same standard on RuboCop as we have for puppet IMO.

iankronquist:
Squashed together.

hkenney:
@hlindberg We are using the same rubocop standards as puppet for strings. There was just some confusion because we don't use rubocop to check anything in the templates directory due to the fact that it's full of .erb files. :)

hkenney:
Okay, so I few suggestions. First, when I test I don't see file or line numbers, I see warnings like this:

```
[warning] Unknown file name or line number: The parameter num_c is documented, but doesn't exist in your code.
```

I also think that for consistency, we should try to make the format of our warnings similar to the way YARD does things, the same warning from YARD looks like this:

```
[warn]: @param tag has unknown parameter name: num_c
    in file `lib/puppet/functions/4x_function.rb' near line 12`
```

The last thing that I think would be ideal (unless this proves to be way too complicated and involve a lot of monkey patching) is to figure out how to silence YARD's warnings about missing parameters. Due to the fact that puppet functions are just a little different from Ruby, YARD will issue warnings for `@param` tags without existing parameters, even with the parameters are there. That means that our warnings are not particularly meaningful because they are mixed in with all the red herrings.

For example with this puppet function:
```
 # When given two numbers, returns the one that is larger.
 # You could have a several line description here if you wanted,
 # but I don't have much to say about this function.
 #
 # @example using two integers
 #   $bigger_int = max(int_one, int_two)
 #
 # @return [Integer] the larger of the two parameters
 #
 # @param num_c [Integer] the first number to be compared
 # @param num_b [Integer] the second number to be compared
 Puppet::Functions.create_function(:max) do
   def max(num_a, num_b)
     num_a >= num_b ? num_a : num_b
   end
 end
```

I see these warnings: 
```
[warn]: @param tag has unknown parameter name: num_c
    in file `lib/puppet/functions/4x_function.rb' near line 12
[warn]: @param tag has unknown parameter name: num_b
    in file `lib/puppet/functions/4x_function.rb' near line 12
[warn]: @param tag has unknown parameter name: num_c
    in file `lib/puppet/functions/4x_function.rb' near line 12
[warning] Unknown file name or line number: The parameter num_c is documented, but doesn't exist in your code.
```

iankronquist:
1. Fixed. I originally based the warning method off of clang because I think clang's warnings are pretty. This was probably unwise.
2. This turned out to be a bug. `<` should have been `>=`.
3. There is no easy way to fix this without monkey patching yard's docstring_transformer or registering our own tag, say puppet_parameter.

hlindberg:
reg 3. Can we perhaps fool yard into believing there is such a parameter?

hlindberg:
Not sure we can completely monkey patch away the yard behavior as it is of value when operating on Ruby code that is included in a module - helper classes etc. Is there some way we can make that conditional depending on if it is one of "our things" being processed or not?

hkenney:
@hlindberg That feels kind of tricky because puppet functions (which is the issue in question at the moment) is Ruby code which is being processed via the Ruby parser. 

We do have our own handlers for puppet functions which inherit from `YARD::Handlers::Ruby::Base`, it may be possible to use those as a distinction somehow? I'm not sure what kind of information we have when we're in `YARD::DocstringParser`. If we have some information about the code object, we should be able to figure out if it is "ours" or not.

@iankronquist What do you think? Did you see any info about code objects when you were in there with the debugger?

iankronquist:
Now with more tests! @hkenney @hlindberg r?

hkenney:
I'm working on trying to figure out how to update strings tests so that they are not tied to rspec 2. That way existing tests and the tests in this PR can coexist. Stay tuned

lib/puppet/face/strings.rb
```diff
@@ -50,6 +50,7 @@ def check_required_features
       # For now, assume the remaining positional args are a list of manifest
       # and ruby files to parse.
       yard_args = (args.empty? ? MODULE_SOURCEFILES : args)
+      class YARD::Logger; def progress(*args); end; end
```
hlindberg: break this up on multiple lines

hkenney: Actually this is for debugging purposes and should be removed entirely. It's gutting a method so that threads do not get in the way of debugging.

spec/unit/puppet_x/puppetlabs/strings/yard/puppet_4x_function_handler_spec.rb
```diff
@@ -54,12 +54,21 @@ def the_namespace()
...
-        This is not ruby code
-    RUBY
-
-    expect(YARD::Registry.all).to be_empty
+  it "should issue a warning if the parameter names do not match the docstring" do
+      expected_output_not_a_param = "[warn]: @param tag has unknown parameter" +
+        " name: not_a_param \n    in file `(stdin)' near line 3"
+      expected_output_also_not_a_param = "[warn]: @param tag has unknown " +
+        "parameter name: also_not_a_param \n    in file `(stdin)' near line 3"
+      expect {
```
hlindberg: the above looks like a mock up with static line number, use of 'near' etc.

lib/puppet/face/strings.rb
hlindberg: ah ok - then it is better to have it as a one-liner :-) if useful later - comment it out and have a comment # For debugging, uncomment to .... or somesuch? I have used that in the parser in a place where it was difficult to remember where to add it back in when doing certain kinds of debugging. - just a though.

hkenney: :+1: Seems like a good idea to me! I use this line often (I have it written on a sticky note)

hkenney: @iankronquist Could you go ahead and make this change? Just comment it out and leave a comment above explaining why it is useful for debugging (e.g. prevent threading so we can use the debugger)

iankronquist: Fixed.

hlindberg:
+1 on merging

PR #32, opened 2015-07-02
(maint) Remove html-matchers dependency

hkenney:
Prior to this commit, some of the unit tests for strings relied on
the rspec-html-matchers gem, which does not seem to get along with
rspec 3. This meant that we had to tie our tests to rspec 2 which
has become too inconvenient.

In order to remove the gem, use simple string matching instead.
Update the Gemfile so that it no longer requires the matchers and
no longer has rspec tied to 2.7.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,10 +14,9 @@ else
 end
 
 group :test do
-  gem "rspec", "~> 2.14.1", :require => false
+  gem "rspec"
   gem 'mocha'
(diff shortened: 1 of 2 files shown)


iankronquist:
+1 Looks good to me.

hkenney:
Hang on, checking to see why Ruby 2.0 is upset in Travis

hkenney:
Travis has been appeased :+1:

PR #34, opened 2015-07-12
Making this module available as a gem and adding rake tasks.

danzilio:
Not sure if I'm the only one who wants this. I have some rake tasks to generate docs with Strings that just shell out and run the commands with Puppet; this is an attempt to do that natively. This is not the most maintainable way to do this, but my goal was to be most minimally invasive. I can take a different, more modular approach if there's interest in this feature.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,9 +1,10 @@
 source 'https://rubygems.org'
 
+gemspec
+
 gem 'yard'
 gem 'rgen'
 gem 'redcarpet'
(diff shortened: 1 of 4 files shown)


hkenney:
Hey @danzilio thanks so much! I've heard from several people that they'd like to see strings as a gem, thanks for getting this moving. We'll take a look at this pull request soon and get back to you.

danzilio:
Resubmitted this as #54

PR #35, opened 2015-07-14
Type checking/pdoc 21

iankronquist:
XKCD: Git Commit

### DIFF
--- a/lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
+++ b/lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
@@ -8,6 +8,23 @@ class PuppetX::PuppetLabs::Strings::YARD::Handlers::HostClassHandler < PuppetX::
         param_tuple << ( a[1].nil? ? nil : a[1].source )
       end
     end
+    tp = Puppet::Pops::Types::TypeParser.new
+    param_type_info = {}
+    statement.pops_obj.parameters.each do |pop_param|
+      # If the parameter's type expression is nil, default to Any
+      if pop_param.type_expr == nil
+        param_type_info[pop_param.name] = Puppet::Pops::Types::TypeFactory.any() 
+      else
+        begin
+          param_type_info[pop_param.name] =  tp.interpret_any(pop_param.type_expr)
+        rescue Puppet::ParseError
+          # If the type could not be interpreted, default to type Any
+          param_type_info[pop_param.name] = "TypeError - " +
+            "#{pop_param.type_expr} isn't a valid type."
+        end
+      end
+    end
+    obj.type_info = [param_type_info]
 
     statement.pops_obj.tap do |o|
       if o.parent_class
--- a/lib/puppet_x/puppetlabs/strings/yard/handlers/puppet_4x_function_handler.rb
+++ b/lib/puppet_x/puppetlabs/strings/yard/handlers/puppet_4x_function_handler.rb
@@ -52,19 +52,24 @@ class Puppet4xFunctionHandler < YARD::Handlers::Ruby::Base
     obj = MethodObject.new(function_namespace, name) do |o|
     end
 
-    # overload_signatures is a array of arrays of tuples or empty:
+    # The data structure for overload_signatures is an array of hashes. Each
+    # hash represents the arguments a single function dispatch (aka overload)
+    # can take.
     # overload_signatures = [
-    #   [ # First function dispatch arguments
+    #   { # First function dispatch arguments
     #     # argument name, argument type
-    #     ['arg0', 'Variant[String,Array[String]]'
-    #     ['arg1', 'Optional[Type]']
-    #   ],
-    #   [ # Second function dispatch arguments
-    #     ['arg0', 'Variant[String,Array[String]]'
-    #     ['arg1', 'Optional[Type]']
-    #     ['arg2', 'Any']
-    #   ]
+    #     'arg0': 'Variant[String,Array[String]]',
+    #     'arg1': 'Optional[Type]'
+    #   },
+    #   { # Second function dispatch arguments
+    #     'arg0': 'Variant[String,Array[String]]',
+    #     'arg1': 'Optional[Type]',
+    #     'arg2': 'Any'
+    #   }
     # ]
+    # Note that the order for arguments to a function doesn't actually matter
+    # because we allow users flexibility when listing their arguments in the
+    # comments.
     overload_signatures = []
     statement.traverse do |node|
       # Find all of the dispatch methods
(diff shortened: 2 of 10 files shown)


iankronquist:
@hkenney r?

iankronquist:
Squashed.

lib/puppet_x/puppetlabs/strings/yard/templates/default/template_helper.rb
```diff
@@ -105,6 +105,106 @@ def comment_only_param_details(param_tags)
...
+    #        "#{param[:name]} parameter is declared as types #{param[:types]} in " +
+    #        "the docstring, but the code specifies the types " +
+    #        "#{actual_types.inspect} Sorry, the file and line number could" +
+    #        "not be determined."
+    #    end
+    #    log.warn warning
+    #  end
+    #end
+  end
+
```
hkenney: Should this be deleted or uncommented...?

spec/unit/puppet_x/puppetlabs/strings/yard/examples/puppet_4x_mismatched_types.rb
```diff
@@ -0,0 +1,12 @@
...
+Puppet::Functions.create_function(:max) do
+  dispatch max_1 do
+    param 'Integer', :num_a
+    param 'Integer', :num_b
+  end
+
+  def max_1(num_a, num_b)
+    num_a >= num_b ? num_a : num_b
+  end
+end
```
hkenney: Let's go ahead and remove this file since it's not being used in a test

lib/puppet_x/puppetlabs/strings/yard/handlers/puppet_4x_function_handler.rb
```diff
@@ -167,7 +178,6 @@ def function_namespace
   # Extracts the Puppet function name and options hash from the parsed
   # definition.
   #
-  # @return [(String, Hash{String => String})]
```
iankronquist: This line probably died in an unfortunate merging accident.

spec/unit/puppet_x/puppetlabs/strings/yard/host_class_handler_spec.rb
```diff
@@ -59,4 +59,5 @@ class puppet_enterprise::mcollective::client::certs { }
 
     expect(namespace).to_not be_nil
   end
+
```
iankronquist: This doesn't need to be here.

spec/unit/puppet_x/puppetlabs/strings/yard/puppet_4x_function_handler_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'puppet/face/strings'
```
iankronquist: neither does this

iankronquist:
@hkenney changed according to your comments/

hkenney:
It looks like we still have an empty file committed? 'spec/fixtures/manifests/site.pp'

iankronquist:
@hkenney fixed

hkenney:
@hlindberg Could you give this a :+1: on the parser stuff?

lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
```diff
@@ -8,6 +8,21 @@ class PuppetX::PuppetLabs::Strings::YARD::Handlers::HostClassHandler < PuppetX::
         param_tuple << ( a[1].nil? ? nil : a[1].source )
       end
     end
+    tp = Puppet::Pops::Types::TypeParser.new
+    param_type_info = []
+    # Ensure that this pop object has parameters
+    if statement.pops_obj.respond_to? :parameters
```
hlindberg: hm, not the best way to ensure this only deals with parameters of things we want to document. Much better to test the type of the pops_obj. Some ruby-ists think "responds_to?" is a great idea and use it judiciously, but it should really be reserved for last resort / meta programming / adapting to versions of an imlementation etc. as it leads to horrible spaghetti and opens for surprising behavior.

lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
```diff
@@ -8,6 +8,21 @@ class PuppetX::PuppetLabs::Strings::YARD::Handlers::HostClassHandler < PuppetX::
...
+      param_type_info = {}
+      statement.pops_obj.parameters.each do |pop_param|
+        begin
+          param_type_info[pop_param.name] =  tp.interpret_any(pop_param.type_expr)
+        rescue Puppet::ParseError
+          # If the type could not be interpreted, default to type Any
+          param_type_info[pop_param.name] =  tp.parse('Any')
```
hlindberg: No need to create a new instance of 'Any' for each untyped parameter - they can all refer to the same Any instance. To  make this work for all puppet versions call Puppet::Pops::Types::TypeFactory.any() instead (it will provide a singleton in newer puppet versions, and do the right thing in older versions).

lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
```diff
@@ -8,6 +8,21 @@ class PuppetX::PuppetLabs::Strings::YARD::Handlers::HostClassHandler < PuppetX::
...
     end
+    tp = Puppet::Pops::Types::TypeParser.new
+    param_type_info = []
+    # Ensure that this pop object has parameters
+    if statement.pops_obj.respond_to? :parameters
+      param_type_info = {}
+      statement.pops_obj.parameters.each do |pop_param|
+        begin
+          param_type_info[pop_param.name] =  tp.interpret_any(pop_param.type_expr)
+        rescue Puppet::ParseError
```
hlindberg: I think a nil type_expr should be handled differently from a broken type_expr - when it is nil, it should default to Any. But if there is an error the documentation should say something else e.g if given `Integer[herp, derp]` - then the result should not be Any - maybe emit something like "TypeError - <message>"

lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
iankronquist: As I think about this again I'm wondering if this is truly necessary at all. I think I was being defensive since some puppet types have parameters and some do not, but this us clearly going to be a HostClassDefinition which always has parameters!

iankronquist:
@hlindberg addressed your comments.

lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
hkenney: Are you sure a HostClassDefintion always has parameters? Maybe I'm confused

hkenney: jk we're sure! It's in the model meta

lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
```diff
@@ -8,6 +8,23 @@ class PuppetX::PuppetLabs::Strings::YARD::Handlers::HostClassHandler < PuppetX::
...
+      # If the parameter's type expression is nil, default to Any
+      if pop_param.type_expr == nil
+        param_type_info[pop_param.name] = Puppet::Pops::Types::TypeFactory.any() 
+      else
+        begin
+          param_type_info[pop_param.name] =  tp.interpret_any(pop_param.type_expr)
+        rescue Puppet::ParseError
+          # If the type could not be interpreted, default to type Any
+          param_type_info[pop_param.name] = "TypeError - " +
+            "#{pop_param.type_expr} isn't a valid type."
```
hlindberg: Did you try this? I am not sure you get human readable output from to_s on any AST object. There is a "label provider" that can give human readable output about an expression, but it probably does not provide any real information about what the problem is to the user. The error message from the TypeParser should be of more value.

iankronquist: On further inspection, interpret_any can't even raise this exception. The interpret method can, and I originally thought that interpret_any called out to interpret. I have looked into what other exceptions it can raise (is there a tool to make this easier?) but nothing seems likely. This is ruby, so potentially anything can be passed down this hole, but I can't defend against absolutely every exception. I'm going to remove this code path altogether.

iankronquist: What I just said is wrong.

PR #36, opened 2015-07-15
Actually/pdoc 19

iankronquist:
@hkenney r?


iankronquist:
FYI this branches off PR #35

hkenney:
Just noting: We'll need to rebase this when #35 is in

PR #37, opened 2015-07-16
(PDOC-30) Fix Markdown parsing lists parsing

iankronquist:
The transformer comment matching regex matched all whitespace after a comment.
Lines which were effectively "blank" and just had a comment would be erased.
Markdown lists need to end with a blank line. This messed up markdown
formatting.

The culprit is the regex /^\s*#\s/ which matches all of the comment ' #\n',
however we want to leave the newline alone for markdown to parse.
We replace the regex with /^\s*#[ \t]/ which will only match tabs or spaces
after the hash and leave our beloved newline alone.

Fixes [PDOC 30 on JIRA](https://tickets.puppetlabs.com/browse/PDOC-30)


hkenney:
I will merge this once we sort out the CI failures :+1:

PR #40, opened 2015-07-20
(maint) Parameters with no type info get Any type

iankronquist:
I'm not sure if this is the right approach and can't really tell if this fixes the problem since I can't reproduce the failure.


hkenney:
Going to go ahead and close this. It's no longer necessary since we're not going to support puppet 3.6.0 anymore.

PR #41, opened 2015-07-27
Nested classes/pdoc 35

iankronquist:


spec/unit/puppet/face_spec.rb
```diff
@@ -79,6 +79,21 @@
...
+            Dir.chdir('test')
+
+            Puppet::Face[:strings, :current].yardoc
+
+            expect(read_html(tmp,
+              'test', 'outer.html')).to include("Puppet Class: outer")
+            expect(read_html(tmp, 'test',
+              'outer/middle.html')).to include("Puppet Class: middle")
+            expect(read_html(tmp, 'test',
+              'outer/middle/inner.html')).to include("Puppet Class: inner")
```
HAIL9000: I think it would be worthwhile to check in the test that the classes have the expected namespaces. I believe the namespace for each class should reflect the class it's nested in.

iankronquist: That description is also incorrect.

iankronquist:
@HAIL9000 added a test to address your comment and fixed test name.

iankronquist:
@HAIL9000 I think Shaigy wants to test this. Would you mind looking at this again?

PR #44, opened 2015-08-03
(PDOC-21) Check mismatched types in defined types

iankronquist:
@HAIL9000 Shaigy pointed this out to me.


HAIL9000:
@iankronquist Could you go ahead and make the commit message a bit more detailed? Otherwise this looks good. Is this already being done for classes?

iankronquist:
Yes this is already being done for classes. I'll fix the commit message.

PR #45, opened 2015-08-04
(maint) Don't print extraneous "true".

iankronquist:
Suppress it by printing an empty string instead.


iankronquist:
@HAIL9000 r?

PR #46, opened 2015-08-07
(PDOC-35) Support types and providers

iankronquist:
~~This is not ready yet, but I thought I'd put it up so people can take a look at it.~~ I still need to:
- [x] Handle generated property methods (`mk_resource_method`)
- [x] ~~Handle default values? Is this a requested feature? I see these often mentioned in the docs.~~
- [x] Write tests.
- [x] Squash this nonsense into a reasonable number of commits.


iankronquist:
I wrote three tests for the providers handler. What else should I test? @HAIL9000 Thoughts?
Also I spoke to Charlie about meta programming madness and am looking into adding a custom tag to yard.

iankronquist:
@HAIL9000 this is ready for review. Would you like me to squash it before or after you review it?

HAIL9000:
@iankronquist Let's wait on squashing in case we need to make more changes. We'll definitely want this broken down into several commits when we squash, so let's wait until it's ready to go so we can organize those in a way that makes sense.

lib/puppet/face/strings.rb
```diff
@@ -55,6 +56,9 @@ def check_required_features
       # all over the terminal. This should definitely not be in real code, but
       # it's very handy for debugging with pry
       #class YARD::Logger; def progress(*args); end; end
+      YARD::Tags::Library.define_directive("puppet.provider.param",
```
HAIL9000: Hmm... not sure how I feel about the name for this directive but I can't think of a better option off the top of my head. May be worth reaching out for suggestions.

iankronquist: It's a bit long but I like the namespacing.

iankronquist:
I talked with Nick. This isn't sufficient.
Here is a sketch of some of what we need:
img_20150826_150036
- [ ] Rename  provider to type
- [ ] *Type*
- [ ] List providers from within type (modify template and type logic)
- [ ]  associate a provider with a type (create a provider handler and search registry for existing type)
- [ ] type features
- [ ] type features desc
- [ ] type parameters
- [ ] param namevar (4 ways of determining -- ick)
- [ ] param desc
- [ ] param default
- [ ] param allowed values
- [ ] prop desc
- [ ] prop default
- [ ] prop allowed values
- [ ] *Providers*
- [ ] prov desc
- [ ] prov commands
- [ ] prov confines
- [ ] prov defaultfor
- [ ] prov features

iankronquist:
This needs a hell of a rebase but is ready for review.

hlindberg:
@nfagerlund ping - are you happy with the changes?

lib/puppet_x/puppetlabs/strings/yard/handlers/provider_handler.rb
```diff
@@ -0,0 +1,94 @@
...
+    defaults = {}
+    do_block = statement.jump(:do_block)
+    do_block.traverse do |node|
+      if is_a_func_call_named? 'desc', node
+        content = node.jump(:tstring_content)
+        # if we found the string content extract its source
+        if content != node
+          # The docstring is either the source stripped of heredoc
+          # annotations or the raw source.
+          if @heredoc_helper.is_heredoc? content.source
```
hlindberg: As a matter of style:
While it is ok to leave out the () around arguments in a method call, I prefer that there are () everywhere except when the methods are designed to be used in a human friendly DSL style. I find that readability increases in long chains of words (esp when there are also underscores). The one above not so bad, but below there is a line that says:
```
docstring = @heredoc_helper.process_heredoc content.source
```
I would prefer that to read:
```
docstring = @heredoc_helper.process_heredoc(content.source)
```

Using parentheses also reduces the risk that someone later makes mistakes if there is the need to continue the expression.

(I see no need to go through all of the code to change this, but if you rework it for other reasons, this is something to consider).

lib/puppet_x/puppetlabs/strings/yard/handlers/provider_handler.rb
```diff
@@ -0,0 +1,94 @@
...
+        node.traverse do |s|
+          if s.type == :assoc
+            k = s.first.jump(:ident).source
+            v = s[1].first.source
+            confines[k] = v
+          end
+        end
+      elsif is_a_func_call_named? 'has_feature', node
+        list = node.jump :list
+        if list != nil and list != node
```
hlindberg: We should not use Ruby `and`, `or`, and `not` because of their peculiar precedence. Always use `&&`, `||` and `!`.

iankronquist: Fixed here and elsewhere in this file. The real solution would be to make another PR with this sort of cleanup everywhere and add a rule to the linter.

nfagerlund:
Sorry for the delay. Yes, I'm happy with what we have! I'm not familiar enough with the code to review it, but the output Ian has shown me is a good first pass at types / providers docs. The foundation is solid enough that we can make further improvements incrementally. I'm :+1: to merge.

iankronquist:
Thanks @nfagerlund @HAIL9000 @hlindberg for your feedback and suggestions.
This has now been rebased and is ready to go once Travis finishes.
@HAIL9000 Could you take a look at #47, the accompanying docs?

iankronquist:
... and a debugging comment creeps in at the last minute.

lib/puppet_x/puppetlabs/strings/yard/code_objects/defined_type_object.rb
```diff
@@ -1,7 +1,22 @@
+require 'json'
+
 class PuppetX::PuppetLabs::Strings::YARD::CodeObjects::DefinedTypeObject < PuppetX::PuppetLabs::Strings::YARD::CodeObjects::PuppetNamespaceObject
   # A list of parameters attached to this class.
   # @return [Array<Array(String, String)>]
   attr_accessor :parameters
   attr_accessor :type_info
 
+  def to_s
+    require 'pry'; binding.pry
```
HAIL9000: Looks like we have a bunch of remaining debug statements we should take out (I think these are what's causing the test failures in Travis)

HAIL9000: jk you already noticed this :+1:

iankronquist:
@HAIL9000 Travis is green.

lib/puppet_x/puppetlabs/strings/yard/templates/default/fulldoc/html/setup.rb
```diff
@@ -27,6 +27,22 @@ def generate_puppet_plugin_list
...
+
+def generate_puppet_provider_list
+  @items = options.objects.select{|o| [:provider].include? o.type} if options.objects
+  @list_title = "Puppet Provider List"
+  @list_type = "puppet_provider"
+  generate_list_contents
+end
+
+
+
```
HAIL9000: Ultra picky comment, but do we need so much white space here?

iankronquist: Fixed whitespace.

ferventcoder:
:+1:

PR #47, opened 2015-08-11
(PDOC-35) Document documenting puppet providers

iankronquist:
Add instructions on how to use the `@!puppet.provider.param` directive to
document parameters which are created via metaprogramming.

@HAIL9000 I'm going to mark this as blocked by #46

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,39 @@ And here's an example of how you might document a class:
      ) { }
 ```
 
+Puppet providers often rely on different types of metaprogramming to create
+parameters and methods. Since Strings does not execute the code which is being
+documented you must provide hints on how to document your code. To document a
+parameter which is automatically created you must use the special directive
+`@!puppet.provider.param` Which may take types, the parameter name, and a
+description.
+
+Strings will automatically extract the `@doc` provider docstring and any `desc`
+parameter docstrings.
+
+```ruby
+# @!puppet.provider.param my_parameter This parameter needs to be explicitly
+# documented
+Puppet::Type.newtype(:minifile) do
+
+  @doc = "Manages files, including their content, ownership, and permissions.
+    The provider can manage symbolic links."
+
+  # This function does some metaprogramming on the new type.
+  mk_resource_methods
+
+  newparam(:path) do
+    desc <<-'EOT'
+      The path to the file to manage.  Must be fully qualified.
+    EOT
+    # ... do stuff here
+  end
+  # ...
+end
+
+
+```
+
 Here are a few other good resources for getting started with documentation:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -158,6 +158,39 @@ And here's an example of how you might document a class:
...
+documented you must provide hints on how to document your code. To document a
+parameter which is automatically created you must use the special directive
+`@!puppet.provider.param` Which may take types, the parameter name, and a
+description.
+
+Strings will automatically extract the `@doc` provider docstring and any `desc`
+parameter docstrings.
+
+```ruby
+# @!puppet.provider.param my_parameter This parameter needs to be explicitly
```
ferventcoder: Would it be better to use `path` here instead of `my_param` since the param below is `path`?

ferventcoder: I was trying to find `my_param` to understand the real world use in the example.

HAIL9000: /cc @iankronquist thoughts? Sounds like this may make the example more readable

iankronquist: Actually that wouldn't be correct here. The directive is for documenting things which strings doesn't pick up on automatically. Strings will document the `path` parameter, but it will not document `my_param` which we are pretending will be generated automatically by metaprogramming.
Does that make sense?

ferventcoder: Right on! I didn't pick up on that from the words above, but now that you said that, I do from rereading it. Perhaps you should mention the second part first. I word smithed this a little to help clarify.

~~~md
Strings will automatically extract the `@doc` provider docstring and any `desc` parameter/property docstrings.

Sometimes however, Puppet providers use metaprogramming to create parameters and methods automatically. In those cases Strings will not be able to document them automatically (Strings doesn't execute the code that would generate those parameters), so you will need to provide hints on how to document your code. To document a parameter which is automatically created you must use the special directive `@!puppet.provider.param` Which may take types, the parameter name, and a description.
~~~

Does  "parameter which is automatically created" also apply to properties that are auto created? I assume yes. I also don't know how often one could create a property automatically as the work involved in those is more difficult.

ferventcoder: Sorry, I fixed up the wording above to make it more readable.

iankronquist: 1. Your wordsmithing is very nice. Would you like top open a PR against the README?
2. I didn't know you could create a property automatically. We should probably add a directive to support that. If you don't mind, could you file a bug in JIRA? If you're really curious about Strings internals I could give you some pointers on how to implement a directive.

Thanks for your help!

ferventcoder: I'm not sure you can create a property automatically haha. It would be a crazy amount of code genning.

ferventcoder: Done - #56.

PR #48, opened 2015-08-12
(PDOC-21) Only create HostClass parameters once

iankronquist:
You can pass a block to the constructor of the hostclass. For some reason Yard
is yielding the block twice. Our code is not yielding the block because we only
use yield in one place which is unrelated. I don't think this is intended
behavior on yard's part, it certainly seems odd, but we can dodge the question
and not have to wait for an upstream patch by extracting the logic from the
block and running it after the object has been initialized.


iankronquist:
@HAIL9000 r?

HAIL9000:
I'm not even going to wait for Travis #YOLO

PR #49, opened 2015-08-17
(PDOC-21) Duplicate warnings

iankronquist:
@HAIL9000 could you test this really well please? Also includes a fair bit of refactoring which needed to happen.

### DIFF
--- a/lib/puppet_x/puppetlabs/strings/yard/templates/default/hostclass/setup.rb
+++ b/lib/puppet_x/puppetlabs/strings/yard/templates/default/hostclass/setup.rb
@@ -3,13 +3,6 @@ include T('default/definedtype')
 def init
   super
   sections.push :subclasses
-
-  @template_helper = TemplateHelper.new
-  @template_helper.check_parameters_match_docs object
-  params = object.parameters.map { |param| param.first }
-  param_tags = object.tags.find_all{ |tag| tag.tag_name == "param"}
-  param_details = @template_helper.extract_param_details(params, param_tags) unless params.nil?
-  @template_helper.check_types_match_docs object, param_details
 end
 
 def subclasses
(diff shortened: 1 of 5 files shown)


lib/puppet_x/puppetlabs/strings/yard/templates/default/hostclass/setup.rb
```diff
@@ -3,13 +3,6 @@
 def init
   super
   sections.push :subclasses
-
-  @template_helper = TemplateHelper.new
-  @template_helper.check_parameters_match_docs object
-  params = object.parameters.map { |param| param.first }
-  param_tags = object.tags.find_all{ |tag| tag.tag_name == "param"}
-  param_details = @template_helper.extract_param_details(params, param_tags) unless params.nil?
-  @template_helper.check_types_match_docs object, param_details
```
iankronquist: This is the meat of the matter. The rest is just tests and refactoring.

spec/unit/module_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+      ensure
+        Dir.chdir(old_dir)
+      end
+    end
+  end
+
+  def self.read_html(dir, modulename, file)
+    File.read(File.join(dir, modulename, 'doc', file))
+  end
+end
```
HAIL9000: So all the functionality for helping with tests should live under spec/lib/strings_spec. If we have any other helper classes floating around they should also be moved in there.

lib/puppet_x/puppetlabs/strings/yard/templates/default/hostclass/setup.rb
HAIL9000: So did we have this functionality in two places causing the duplication? All I see is this removal and then the spec updates, so I'm guessing this was already being handled somewhere else?

iankronquist:
@HAIL9000 moved.

HAIL9000:
Also before this is ready for merge we should rebase down to just a couple commits that all have the (PDOC-21) reference in the title.

iankronquist:
@HAIL9000 here it is, nicely rebased.

PR #50, opened 2015-08-20
(PDOC-38) Prevent warnings from being issued twice

iankronquist:


iankronquist:
Rebased off of #46. This is, unfortunately, the best way I can think of doing it.

iankronquist:
Of course I will need to rebase these to make the commits more concise and give them more details, but this seems like it's ready to go. @HAIL9000 @hlindberg r?

iankronquist:
Rebased.

PR #51, opened 2015-08-31
(MAINT) Add a space between a parameter name type and its description

roidelapluie:
Before this commit:

    packages (Array[String])- A list of packages to be installed

With this commit:

    packages (Array[String]) - A list of packages to be installed


iankronquist:
+1. Thanks for the fix! If you have any other thoughts on how we can improve strings please let us know here or on JIRA.

roidelapluie:
I will. See PDOC-43 and #52.

iankronquist:
I'm going to go ahead and merge this. The people who would also review this are either out or too busy to review this change.

PR #55, opened 2015-09-04
(PDOC-23) Emit json

iankronquist:
Addresses [PDOC-23](https://tickets.puppetlabs.com/browse/PDOC-23).
Nick had some ideas about its current form and this will need to be extended once #46 is merged. I'm putting it up here anticipating our meeting Tuesday the 8th.

### DIFF
--- a/lib/puppet_x/puppetlabs/strings/yard/code_objects/method_object.rb
+++ b/lib/puppet_x/puppetlabs/strings/yard/code_objects/method_object.rb
@@ -13,12 +13,15 @@ class YARD::CodeObjects::MethodObject
   def to_json(*a)
     if self[:puppet_4x_function]
       {
-        "name"                => @name,
-        "file"                => file,
-        "line"                => line,
-        "puppet_version"      => 4,
-        "docstring"           => Puppet::Util::Docs.scrub(@docstring),
-        "documented_params"   => @parameters.map do |tuple|
+        "name"                  => @name,
+        "file"                  => file,
+        "line"                  => line,
+        "function_api_version"  => 4,
+        "docstring"             => Puppet::Util::Docs.scrub(@docstring),
+        "examples"              => self.tags.map do |tag|
+            tag.text if tag.tag_name == 'example'
+          end.compact,
+        "documented_params"     => @parameters.map do |tuple|
           {
             "name" => tuple[0],
             "type" => tuple[1],
@@ -35,17 +38,20 @@ class YARD::CodeObjects::MethodObject
       }.to_json(*a)
     elsif self[:puppet_3x_function]
       {
-        "name"                => @name,
-        "file"                => file,
-        "line"                => line,
-        "puppet_version"      => 3,
-        "docstring"           => Puppet::Util::Docs.scrub(@docstring),
-        "documented_params"   => @parameters.map do |tuple|
+        "name"                  => @name,
+        "file"                  => file,
+        "line"                  => line,
+        "function_api_version"  => 3,
+        "docstring"             => Puppet::Util::Docs.scrub(@docstring),
+        "documented_params"     => @parameters.map do |tuple|
           {
             "name" => tuple[0],
             "type" => tuple[1],
           }
         end,
+        "examples"              => self.tags.map do |tag|
+            tag.text if tag.tag_name == 'example'
+          end.compact,
       }.to_json(*a)
     else
       super
--- a/lib/puppet_x/puppetlabs/strings/yard/code_objects/method_object.rb
+++ b/lib/puppet_x/puppetlabs/strings/yard/code_objects/method_object.rb
@@ -1,0 +1,56 @@
+class YARD::CodeObjects::MethodObject
+
+  # Override to_s and to_json methods in Yard's MethodObject so that they
+  # return output formatted as I like for puppet 3x and 4x methods.
+  def to_s
+    if self[:puppet_4x_function] || self[:puppet_3x_function]
+      name.to_s
+    else
+      super
+    end
+  end
+
+  def to_json(*a)
+    if self[:puppet_4x_function]
+      {
+        "name"                => @name,
+        "file"                => file,
+        "line"                => line,
+        "puppet_version"      => 4,
+        "docstring"           => Puppet::Util::Docs.scrub(@docstring),
+        "documented_params"   => @parameters.map do |tuple|
+          {
+            "name" => tuple[0],
+            "type" => tuple[1],
+          }
+        end,
+       "signatures"       => @type_info.map do |signature|
+          signature.map do |key, value|
+            {
+              "name" => key,
+              "type" => value,
+            }
+          end
+        end,
+      }.to_json(*a)
+    elsif self[:puppet_3x_function]
+      {
+        "name"                => @name,
+        "file"                => file,
+        "line"                => line,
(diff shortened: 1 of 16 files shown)


lib/puppet_x/puppetlabs/strings/yard/code_objects/method_object.rb
```diff
@@ -0,0 +1,34 @@
+class YARD::CodeObjects::MethodObject
```
iankronquist: As opposed to patching Yard's `MethodObject` we could just create `Puppet4xFunctionObject` and `Puppet3xFunctionObject`s which inherit from it. There is some logic identifying puppet and ruby methods (`:method`) right now which would need to be modified as a result.

iankronquist:
Here's a sample of what this emits right now:
```json
{
    "Puppet3xFunctions": {
        "code_object_type": "HostClassObject",
        "docstring": "",
        "parameters": null,
        "type_info": null
    },
    "Puppet4xFunctions": {
        "code_object_type": "HostClassObject",
        "docstring": "",
        "parameters": null,
        "type_info": null
    },
    "foo": {
        "code_object_type": "HostClassObject",
        "docstring": "",
        "parameters": [
            [
                "ident",
                "\"Bob\""
            ],
            [
                "age",
                "10"
            ]
        ],
        "type_info": [
            {
                "age": "Myint",
                "ident": "Mytype"
            }
        ]
    },

}

```

lib/puppet_x/puppetlabs/strings/yard/json_registry_store.rb
```diff
@@ -0,0 +1,69 @@
...
+
+        schema[:defined_types] += obj.select do |key, val|
+          val.type == :definedtype
+        end.values
+        @serializer.serialize(schema.to_json)
+    end
+  end
+
+  # Override the serializer because it puts the data at a whacky path and, more
+  # importantly, mashals the data with a bunch of non-printable characters.
```
iankronquist: Should be 'marshals'

iankronquist:
Rebased to bring it up to date.

iankronquist:
What else does this PR need? I think I followed all Nick's advice on the output structure, but it might be nice if he took one more look at this. It also needs to be squashed.

HAIL9000:
@iankronquist have you had a chance to look into these spec failures?

iankronquist:
Oh dear, I didn't realize there were spec failures. I think I goofed some refactoring in 12f2af6. I'll take a look.

iankronquist:
It looks like I made changes and forgot to update the name of the object I was interested in in the test. It should work now.

HAIL9000:
Awesome, thanks @iankronquist! @nfagerlund @adreyer, if either of you have a chance to give this another once over that would be amazing. Trying to prevent this work from slipping through the cracks since we're getting so close

nfagerlund:
I've got this installed and working, and am playing with it now. It works!

- It should be possible to get JSON on stdout (or a file of our choice) via a command line option or a separate action; having to generate the yard site in order to get the json isn't ideal for our use cases. 
- The `@example` values for 4.x functions aren't surviving into the JSON! This looks like a bug. 
- Still a little wobbly on the name of the `puppet_version` key for functions, but.... also not super fussed about it. Maybe `function_api_version` or something. But, eh!!! I won't hold up the PR on that!

I still like the format, I think we settled on something very usable. The remaining things holding me up from ripping out a bunch of garbage custom code in favor of Strings are mostly either bugs in the parsing that we'll identify over time, or shortcomings in the inputs I'm feeding it (gotta go through the types and use `@-directives` to deal with some of the un-parseable meta-programming, and might need some more kinds of directives to help do that).

nfagerlund:
Ah, one more thing! We should have the JSON format documented somewhere, and ideally use a schema to enforce it! I think Alex and I will both want that before we can justify pushing this into production use, as a guarantee that it'll be less brittle than the garbage it's replacing.

iankronquist:
@nfagerlund excellent

1. Regarding json to stdout, @HAIL9000 what is the status of strings command line flags? I know we currently pass flags to yard via `.yardopts` files. Is it reasonable to create a new set of command line flags for strings?
2. We completely ignore `@example` and all other tags right now. This should be easy to fix I think.
3. Trivial to fix. Again, if you have a better idea how to denote this, by all means let me know.
4. It's documented in the issue, but I think that description is out of date now. I'll go see if I can update the issue. @HAIL9000 where should I put the documentation for the schema? I assume a markdown file somewhere in the repo would be best.

iankronquist:
@nfagerlund do you want the examples just for Puppet 4x functions, for 3x and 4x functions, or for all code?

nfagerlund:
@iankronquist I'm really leaning towards "everything." If there's a convention of not using it, like in the 3.x functions, that'll stay fine, but I can see people wanting to use these in classes, custom resource types, etc. Let's make it a first class citizen everywhere. 

Also: If you use an example in, say, a parameter docstring (for a resource type), it should probably attach to the parameter rather than the parent resource type.

iankronquist:
@nfagerlund Done.

iankronquist:
Okay I have now implemented two command line flags:
* `--emit-json $FILE`
* `--emit-json-stdout`

nfagerlund:
@iankronquist Sweet. I've found three small problems:

* Both options are undocumented. I can't remember how to do that with a face, but I remember it was pretty simple.
* `--emit-json` can't seem to write to files that aren't in the cwd. I tired `~/Desktop/stdlib.json` and `/Users/nick/Desktop/stdlib.json`, and neither worked. 
* `--emit-json-stdout` adds some junk after the JSON data: 

```
Files:         115
Modules:         5 (    3 undocumented)
Classes:         1 (    0 undocumented)
Constants:       0 (    0 undocumented)
Methods:       123 (   14 undocumented)
Puppet Classes:     2 (    0 undocumented)
Puppet Defined Types:     0 (    0 undocumented)
Puppet Types:     2 (    2 undocumented)
Puppet Providers:     1 (    1 undocumented)
 85.07% documented
```

iankronquist:
@nfagerlund All fixed.

nfagerlund:
@iankronquist Rockin. Just tried it and it looks good. This could very well be good to go. @HAIL9000?

lib/puppet_x/puppetlabs/strings/yard/code_objects/method_object.rb
HAIL9000: Would you mind filing a ticket for this improvement so we don't lose track of it?

HAIL9000:
This all looks :+1: to me! @iankronquist would you mind updating the README to reflect this new functionality (including a description of the flags and how they can be used, and probably a reference to the JSON schema) and then squashing down to fewer commits? Then I think we can merge this.

iankronquist:
@HAIL9000 Updated README and squashed down to 4 commits.

lib/puppet_x/puppetlabs/strings/yard/code_objects/method_object.rb
iankronquist: [PDOC-64](https://tickets.puppetlabs.com/browse/PDOC-64) has been filed.

iankronquist:
Yay! Thanks @nfagerlund and @HAIL9000 for your help finishing this up.

PR #58, opened 2015-09-14
(maint) Update changelog for 0.3.0 release

iankronquist:
This assumes that #55 will be merged before the release date of next Tuesday the 21st of September.


HAIL9000:
I think there is definitely some extranious information in here that is not needed. If you need an example of how we usually do CHANGELOGs for the puppet modules, check out one of the other puppetlabs modules (like puppetlabs-apache for example): https://github.com/puppetlabs/puppetlabs-apache/blob/master/CHANGELOG.md

HAIL9000:
(also since these modules are owned by Puppet Labs I don't think we generally mention individual contributors in the CHANGELOGs)

iankronquist:
@HAIL9000 I pared it down a bit. Let me know what you think.

PR #59, opened 2015-09-14
(PDOC-35) Format generated html properly

iankronquist:
Shaigy Nixon on Jira:
"Just a trivial thing. Looks like one line break is missing from the type's
output html between the property and the allowed values for parameter. Rest all
looks great."
https://tickets.puppetlabs.com/browse/PDOC-35?focusedCommentId=215400&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-215400

As an aside, I would really like to rewrite this method to be a real erb
template and not this nonsense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ Installing the Yard Gem
 -----------------------
 
 **Installing the Yard Gem with Puppet**
+
+
 The easiest way to install the Yard gem is with Puppet itself.
 
 For Puppet 4.x:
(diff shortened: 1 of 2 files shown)


HAIL9000:
If you want to add another erb template you can definitely do that. But if you think there is further work that needs to be done that this pull request won't cover then you should file a ticket for it so that it won't be lost.

iankronquist:
@HAIL9000 I already have filed a ticket for that, see [PDOC 42](https://tickets.puppetlabs.com/browse/PDOC-42).

PR #60, opened 2015-09-14
Types and providers fixes

iankronquist:
Various fixes Shaigy found while doing QA. Thanks Shaigy!


HAIL9000:
So these need to reference a real ticket. Is there an existing ticket that makes sense to reference?

HAIL9000:
There also seem to be two commits titled "Handle directives with no types" should those be merged?

iankronquist:
Looks like I screwed up the rebase. Here it is again.

PR #66, opened 2015-10-19
(maint) Make some small updates to the README

HAIL9000:
Update the README so that it is accurate

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ It can also print the JSON to stdout:
 $ puppet strings some_manifest.pp --emit-json-stdout
 ```
 
-The schema for the JSON which Strings emits is [well documented](json_dom.md).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,7 +102,7 @@ It can also print the JSON to stdout:
 $ puppet strings some_manifest.pp --emit-json-stdout
 ```
 
-The schema for the JSON which Strings emits is [well documented](json_dom.md).
+The schema for the JSON which Strings emits is [well documented](https://github.com/puppetlabs/puppetlabs-strings/blob/master/json_dom.md).
```
iankronquist: Oops, sorry, should have tested this. I just assumed a relative link would work.

iankronquist:
Looks good to me.

HAIL9000:
It does work on GitHub, but I wasn't sure about the forge. Better safe than sorry!

README.md
HAIL9000: It does work on GitHub, but I wasn't sure about the forge. Better safe than sorry!

PR #68, opened 2015-10-26
(maint) Delete .rb file that was probably commited by accident

roidelapluie:
Extension of the file is .rb but should be .erb, so I guess thir file
was committed by accident.


iankronquist:
It seems to be used in the [provider template]( https://github.com/puppetlabs/puppetlabs-strings/blob/master/lib/puppet_x/puppetlabs/strings/yard/templates/default/provider/setup.rb#L21).

roidelapluie:
@iankronquist the .erb, not the .rb

roidelapluie:
https://github.com/puppetlabhttps://github.com/puppetlabs/puppetlabs-strings/blob/26a9d73ef2b7b9897f9c61a4d82e0cdd9c32b14c/lib/puppet_x/puppetlabs/strings/yard/templates/default/provider/html/command_details.rbs/puppetlabs-strings/blob/26a9d73ef2b7b9897f9c61a4d82e0cdd9c32b14c/lib/puppet_x/puppetlabs/strings/yard/templates/default/provider/html/command_details.erb

iankronquist:
Ah! Yes, sorry, I didn't notice that. Yeah, this is fine then. @HAIL9000 r?

HAIL9000:
@iankronquist It looks the .erb and .rb files have different content. Is HTML from this .rb actually not needed...?

iankronquist:
@HAIL9000 I think the erb is correct, and the rb is wrong. Oops, sorry.

roidelapluie:
I found this because our CI check does:

```
if [[ -d lib ]]
then
echo
echo Checking RB files
echo

find lib -name '*.rb' -print0 | xargs -0 -L 1 ruby -c
fi
```

HAIL9000:
@iankronquist I think I'm still confused... so we don't need the parameter summary or features? Or are those just covered somewhere else that I'm missing

roidelapluie:
@HAIL9000 https://github.com/puppetlabs/puppetlabs-strings/blob/5206c8c62e91b564ae1de1341c662a21b5740bb2/lib/puppet_x/puppetlabs/strings/yard/templates/default/type/html/parameter_details.erb

PR #69, opened 2015-12-08
(maint) Fix ruby namespacing issue

HAIL9000:
Prior to this commit, strings would raise an error when the server
command was run against a module directory that contained the strings
module. This was due to the fact that we were trying to reference
the YARD namespace but our own namespace was being prepending to the
front, making Ruby think the two class were the same.

In order to stop this from happening, prepend a :: to the front of the
YARD version of the class to tell Ruby that we mean the YARD namespace.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ Gemfile.lock
 
 ## YARD
 .yardoc
+.yardwarns
 
 ## MODULE BUILDS
 **/pkg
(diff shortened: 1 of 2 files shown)


iankronquist:
Excellent. I can confirm this works for me!

HAIL9000:
Thanks for trying it out @iankronquist! :D

HAIL9000:
@hlindberg, when you get a chance could you merge this? Guessing it just fell off the radar

PR #70, opened 2015-12-16
(MAINT) Correct the link and task names for the Rake task

garethr:


iankronquist:
I checked and this is correct. Thanks!

HAIL9000:
Merged, thank you @garethr!

PR #71, opened 2016-02-03
Fix issue running strings:generate without a yardopts file

garethr:
Without a .yardopts options array.pop is nil, and everything explodes
with an exception. Applying this patch allowed for running rake
strings:generate cleanly.

Note that this did not affect strings:serve.


HAIL9000:
Looks great, thank you @garethr! :+1:

PR #72, opened 2016-02-03
(PDOC-70) Always generate the JSON doc when running the rake task

garethr:
I'm super enthusiastic about the potential for folks to build things on
top of the JSON output, we've had several related conversations at
Config Management Camp and on puppet-dev recently. I think this is a
sane default, and calling the file strings.json seems a nice
reinforcement.

### DIFF
--- a/lib/puppet-strings/rake_tasks.rb
+++ b/lib/puppet-strings/rake_tasks.rb
@@ -6,7 +6,9 @@ require 'puppet_x/puppetlabs/strings/util'
 namespace :strings do
   desc 'Generate Puppet documentation with YARD.'
   task :generate do
-    PuppetX::PuppetLabs::Strings::Util.generate
+    PuppetX::PuppetLabs::Strings::Util.generate([
+      {emit_json: 'strings.json'}
+    ])
   end
 
   desc 'Serve YARD documentation for modules.'


vStone:
Maybe we need to parse command line arguments in the rake task so we can control this easily when running from a CI (jenkins) server.

garethr:
@vStone I think this being configurable is a good idea, but I also think this is a better default until such a time as things are configurable. I don't at least see a downside of also outputting this file (locally or in a CI system) that can't be avoided by .gitignore or similar, but might be missing something.

hlindberg:
I missed this - is there a ticket for this?

garethr:
@hlindberg not that I created or am aware of. Happy to open one to discuss there if preferable.

hlindberg:
@garethr Please do file a ticket. Tickets and discussions are far more visible and helps with priorities.

garethr:
@hlindberg done, opened https://tickets.puppetlabs.com/browse/PDOC-70

hlindberg:
thanks @garethr

HAIL9000:
@garethr Sorry to be pedantic, but could you also update the commit to reference the ticket? :)

garethr:
@HAIL9000 no problem. Done.

HAIL9000:
@garethr As per the discussion on the Jira ticket, could you quickly add a few sentences in the Rake Tasks section of the README explaining that a .json file will be generated?

HAIL9000:
I'm gonna go ahead and merge this one, then I'll open another pull request to update the README before we release 0.4.0

PR #73, opened 2016-03-11
(maint) Make the json schema clearer

HAIL9000:
Update a few things in the json schema to clear up a few things
that were ambiguous.


HAIL9000:
Please don't merge this yet! I'm not actually sure if I'm right that it has to be done per-file

HAIL9000:
Just kidding I was 100% wrong about this not running against multiple files so I'm going to close this

PR #74, opened 2016-03-16
(PDOC-74) Deprecate module in favor of gem

HAIL9000:
Since it's been agreed that strings is better distributed as a gem, 
do the necessary updates so that we can do a final deprecated release 
of the module in conjunction with the first release of the gem. 

This will also give us a vehicle to release the last of the work @iankronquist 
did this summer and get the JSON generations into the hands of users to try
it out.

funny-gif-girl-party-mad-breaking-things

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+**PLEASE NOTE that the puppetlabs-strings module is being deprecated in favor of a gem. 0.4.0 will be the last release of
+the puppet module and the first release of the Ruby gem. Please see the installation instructions below.**
+
 Puppet Strings
 =============
 [![Build Status](https://travis-ci.org/puppetlabs/puppetlabs-strings.png?branch=master)](https://travis-ci.org/puppetlabs/puppetlabs-strings)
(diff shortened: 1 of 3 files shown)


HAIL9000:
Dang I forgot to update the CHANGLOG, I'll add a commit momentarily

HAIL9000:
Hmmmm not sure why the spec tests are failing, they're passing locally for me

HAIL9000:
ping @kylog @hlindberg for review. I'm trying to figure out what's wrong with Travis and fixing the CHANGELOG

kylog:
LGTM

Re the travis thing you *may* (wild guess) need to do something like this: https://github.com/puppetlabs/puppet/commit/d02e826eca697c56903202a0966ac5a83aebb444

hlindberg:
Looks like bona fide errors; calling `map` on `nil` + more. There are also deprecation warnings from rspec.

HAIL9000:
Yeah I seem to get the deprecation warnings when I run it locally, but not the failures which makes it hard to debug. I'll have to figure out what I have that's different than Travis

HAIL9000:
It's finally green @kylog @hlindberg! \o/

PR #76, opened 2016-03-22
(maint) Don't install beaker for spec tests in travis

HAIL9000:
Prior to this commit, the beaker gem and all its dependencies were
being installed just to run spec tests. Excluding beaker when running
spec tests should make things go faster because we won't have to
install so many gems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 sudo: false
-bundler_args: --without development
+bundler_args: --without development acceptance
 script:
   - "bundle exec rake $CHECK"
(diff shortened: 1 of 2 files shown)


HAIL9000:
The Travis CI specs should still fail, so this isn't ready for merge yet. But I'm curious to see which gems are installed excluding beaker and its dependencies in hopes that it will help narrow down the gem that's causing the spec failure (thanks for the idea @kylog!).

kylog:
:+1:

HAIL9000:
This one's ready to go too!

PR #77, opened 2016-03-24
(PDOC-75) Work with both versions of 'interpret_any'

HAIL9000:
Prior to this commit, strings would fail with puppet 4.4.0 and
newer. This was because strings was making use of a method that was
marked API private and thus subject to change. The method was changed
in 4.4.0 release of puppet to take two variables, meaning that strings
much adjust how it calls the method based on the version of puppet
it's running with.

### DIFF
--- a/lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
+++ b/lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
@@ -16,7 +16,13 @@ class PuppetX::PuppetLabs::Strings::YARD::Handlers::HostClassHandler < PuppetX::
         param_type_info[pop_param.name] = Puppet::Pops::Types::TypeFactory.any()
       else
         begin
-          param_type_info[pop_param.name] =  tp.interpret_any(pop_param.type_expr)
+          # This is a bit of a hack because we were using a method that was previously
+          # API private. See PDOC-75 for more details
+          if Puppet::Pops::Types::TypeParser.instance_method(:interpret_any).arity == 2
+            param_type_info[pop_param.name] = tp.interpret_any(pop_param.type_expr, nil)
+          else
+            param_type_info[pop_param.name] = tp.interpret_any(pop_param.type_exp)
+          end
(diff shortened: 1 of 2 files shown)


lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
```diff
@@ -16,7 +16,13 @@ class PuppetX::PuppetLabs::Strings::YARD::Handlers::HostClassHandler < PuppetX::
         param_type_info[pop_param.name] = Puppet::Pops::Types::TypeFactory.any()
       else
         begin
-          param_type_info[pop_param.name] =  tp.interpret_any(pop_param.type_expr)
+          # This is a bit of a hack because we were using a method that was previously
+          # API private. See PDOC-75 for more details
+          if Puppet.version.start_with?('4.4')
```
kylog: Should we make this `tp.interpret_any.arity == 2`? Failing that can we do a semver comparison so we don't have to bump for each Y release? /cc @hlindberg

HAIL9000: Er you're right, I wasn't thinking of 4.5 and beyond when I wrote this sorry about that

hlindberg: It does not have to be bumped, it was changed just once. It should however not ned to compute it over and over again as the runtime will remain the same the entire run.
It would also be fine to use `tp.method(:interpret_any) == 2` as the predicate (which should also be cached).

it could be done like this:
```
if Puppet::Pops::Types::TypeParser.instance_method(:interpret_any).arity == 2
  def interpret(expr)
    tp.interpret_any(expr, nil)
  end
else
  def interpret(expr)
    tp.interpret_any(expr)
  end
end
```
i.e. conditionally add a method that is suitable for the puppet version. Then call the `interpret(expr)` method instead of calling the method on tp (on line 22 an 24) in this PR

HAIL9000:
@hlindberg @kylog Updated and tests are passing!

PR #80, opened 2016-04-06
Add Puppet type name in front of the provider name

dmitryilyin:
It will prevent providers of different types but with
the same name from merging together.
For example:
type1:ruby
type2:ruby
type3:ruby
Will be different providers instead of the one "ruby" provider.


dmitryilyin:
ping

dmitryilyin:
hello?

HAIL9000:
@dmitryilyin Thanks for the contribution! I just need to find a few minutes to test this out and then I can get it merged

HAIL9000:
Tested this on the puppetlabs/apache module, looks great! 

Thanks again!

PR #82, opened 2016-04-22
(maint) Remove module related content

HAIL9000:
Since we're now transition to strings as a RubyGem only, remove the
module related content from the repository.


kylog:
I'm guessing that including the `puppet-strings-0.4.0.gem` file in this commit was unintentional?

HAIL9000:
@kylog Whoops, you're totally right. I git added too ambitiously, let me remove that.

HAIL9000:
Updated! 👍

PR #83, opened 2016-04-25
(maint) Require rspec 3

HAIL9000:
Prior to this commit, the Gemfile did not specify an rspec version
but relied on methods that were only in rspec 3. In order to prevent
failures when running with versions older than 3, tie the specs to
3.1 or greater, which is what puppet does.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,7 @@ else
 end
 
 group :test do
-  gem "rspec"
+  gem "rspec", "~> 3.1"
   gem 'mocha'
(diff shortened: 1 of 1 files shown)


branan:
I'm :+1: On this. RSpec 2 is ancient, and if even Puppet requires 3 at this point I think we can start bumping at-well in other projects

kylog:
Oh yeah, puppet bumped to rspec 3 over a year ago. Definitely 👍

PR #85, opened 2016-05-02
(maint) Update acceptance tests to use gem

HAIL9000:
Prior to this commit, the acceptance tests were trying to install
strings as a module and failing due to the fact that strings has
transition to a gem and does not have metadata. Update the tests
so that they install strings as a gem instead.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -18,7 +18,7 @@ task :validate do
 end
 
 task :acceptance do
-  sh "puppet module build"
+  sh "gem build puppet-strings.gemspec"
   sh "puppet module build spec/unit/puppet/examples/test"
   sh "BEAKER_set=#{ENV["platform"]} rspec spec/acceptance/*.rb"
 end
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -14,8 +14,8 @@ RSpec.configure do |c|
   c.before :suite do
 
   hosts.each do |host|
-      scp_to(host, Dir.glob('pkg/puppetlabs-strings*.gz').first, 'strings.tar.gz')
-      on host, puppet('module', 'install', 'strings.tar.gz')
+      scp_to(host, Dir.glob('puppet-strings*.gem').first, 'puppet-strings.gem')
+      on host, 'gem install puppet-strings.gem'
 
       scp_to(host, Dir.glob('spec/unit/puppet/examples/test/pkg/username-test*.gz').first, 'test.tar.gz')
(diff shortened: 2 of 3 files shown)


whopper:
👍 looks good to me, will merge after travis is done.

Rakefile
```diff
@@ -18,7 +18,7 @@ task :validate do
 end
 
 task :acceptance do
-  sh "puppet module build"
+  sh "gem build puppet-strings.gemspec"
```
er0ck: @zreichert we should move rototiller to use this method of install, instead of scp the source

spec/acceptance/nodesets/fedora21.yml
```diff
@@ -2,9 +2,9 @@ HOSTS:
   master:
     roles:
       - master
-    platform: fedora-20-x86_64
+    platform: fedora-21-x86_64
```
kylog: Since fedora 21 is EOL maybe bump to fedora 23? Also, I don't know where strings is tested in jenkins, but does this need a matching ci-job-configs change?

HAIL9000: Good call, I'll bump it. It was originally configured to use fedora 20 so chances are the jenkins job is doing this too :\ I'll take a look.

spec/acceptance/nodesets/fedora32.yml
```diff
@@ -2,9 +2,9 @@ HOSTS:
   master:
     roles:
       - master
-    platform: fedora-20-x86_64
+    platform: fedora-23-x86_64
     hypervisor: vcloud
-    template: Delivery/Quality Assurance/Templates/vCloud/fedora-20-x86_64
+    template: Delivery/Quality Assurance/Templates/vCloud/fedora-23-x86_64
```
whopper: @HAIL9000 looks like you typo'd the filename - s/32/23/

HAIL9000: My bad! Updated!

spec/spec_helper_acceptance.rb
```diff
@@ -14,8 +14,8 @@
   c.before :suite do
 
   hosts.each do |host|
-      scp_to(host, Dir.glob('pkg/puppetlabs-strings*.gz').first, 'strings.tar.gz')
-      on host, puppet('module', 'install', 'strings.tar.gz')
+      scp_to(host, Dir.glob('puppet-strings*.gem').first, 'puppet-strings.gem')
```
branan: Does this work correctly if you have a dirty working directory with an old gem in it?

PR #86, opened 2016-05-04
PDOC-77: configurable rake task 

vStone:


README.md
```diff
@@ -205,6 +205,16 @@ To use the rake tasks, `require puppet-strings/rake_tasks` in your `Rakefile`:
...
+If you need a more configurable version of the rake task, you can use `puppet-strings/rake_tasks/generate`.
+
+```ruby
+PuppetStrings::RakeTasks::Generate.new(:documentation) do |task|
+  task.paths = ['site/roles','site/profiles','modules/internal']
+  task.ignore = ['/vendor/','/example/']
+  task.options = {} # disables the strings.json output
+end
+```
+
```
hlindberg: Sorry, but I don't understand how this works - the names are very similar. It says "If you need a more configurable version of the rake task, you can use `puppet-strings/rake_tasks/generate`" as opposed to what? Also, does this alter the meaning of the normal generate ? (I did not quite have the time to read up in detail how the pieces fit together). I think a bit more explanation of how it is supposed to work / be used would be helpful.

hlindberg:
Picky detail, we like the commits to start with the ticket nbr (we have scripts that correlate all commits to tickets at release time). that is, start the commits here with the text (PDOC-77)

vStone:
Fixed documentation to be a bit more verbose, fixed commit messages.

README.md
```diff
@@ -205,6 +205,18 @@ To use the rake tasks, `require puppet-strings/rake_tasks` in your `Rakefile`:
 require 'puppet-strings/rake_tasks'
 ```
 
+By default, the rake task `strings:generate` will scan the manifests and lib directory from your single module. If you need to document a complete, or part of a, puppet tree, you can use `puppet-strings/rake_tasks/generate`. This allows you to set the paths and optionally patterns to include.
+
```
hlindberg: Sorry, still not clear to me. Would I use this instead of `strings::generate`, or do I use this to set the paths and optional patterns to include so that the existing generate uses those options.
Perhaps my confusion comes from ignorance of how you use `strrings::generate` in the first place - but if I am confused others are probably as well.

(Think "optionally patterns to include" reads better as "optionally the patterns to include").

PR #89, opened 2016-06-20
(QENG-3888) Remove static nodeset, use beaker-hostgenerator

smcelmurry:
This is additional work for QENG-3888 in ci-job-configs. We were using static nodesets within spec/acceptance/nodesets and this commit will use beaker-hostgenerator within the Rakefile to get the specified platform .yml configurations automatically. Still defaults to same configuration if none are set before running.


smcelmurry:
@HAIL9000

HAIL9000:
Merged! Thank you @smcelmurry!

PR #90, opened 2016-06-23
(PDOC-71) Workaround for spurious error failures

trevor-vaughan:
This patch catches and reports on issues that have occurred during YARD
processing of the data to be processed.

The issue itself is not resolved but this does provide you with as much
documentation as can be reasonably generated for your materials without
crashing the build.

This was the result of the Baltimore Puppet Users Group Meetup of 22
June 2016.

### DIFF
--- a/lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
+++ b/lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
@@ -2,41 +2,54 @@ class PuppetX::PuppetLabs::Strings::YARD::Handlers::HostClassHandler < PuppetX::
   handles HostClassDefinition
 
   process do
-    obj = HostClassObject.new(:root, statement.pops_obj.name)
+    begin
+      obj = HostClassObject.new(:root, statement.pops_obj.name)
 
-    obj.parameters = statement.parameters.map do |a|
-      param_tuple = [a[0].pops_obj.name]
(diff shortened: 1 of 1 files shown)


HAIL9000:
@hlindberg How would you feel about merging this into strings? It's a complete hack, but at least will stop strings from dying on that exception and allow it to produce as much documentation as possible

lib/puppet_x/puppetlabs/strings/yard/handlers/host_class_handler.rb
```diff
@@ -2,41 +2,51 @@ class PuppetX::PuppetLabs::Strings::YARD::Handlers::HostClassHandler < PuppetX::
...
+      statement.pops_obj.tap do |o|
+        if o.parent_class
+          obj.parent_class = P(:root, o.parent_class)
+        end
       end
-    end
 
-    register obj
+      register obj
+    rescue Exception => e
```
hlindberg: This is too wide - `Exception` includes such nasty things as "out of memory" and other disastrous errors. Should probably use `StandardError` as the process should fail on out of memory and the like.

hlindberg:
This is good to merge once the `Exception` class is changed to `StandardError`.

trevor-vaughan:
@hlindberg @HAIL9000 PR has been updated to catch StandardError and SystemStackError, the latter being what was destroying the parsing process.

PR #92, opened 2016-07-21
(maint) Update YARD to latest version

HAIL9000:
Prior to this commit, puppet strings tests were failing against YARD
versions 0.9.0 and newer. This was due to 1) a change in the YARD
results output and 2) the upstream patch we submitted in PDOC-25
being released.

The upstream patch we submitted for that ticket added a regular
expression that we could override when matching the names of puppet
classes and defines since they begin with lowercase letters (unlike
Ruby). Until this was released with the new YARD gem, we were using
a different method of fixing this issue. Once the gem was out, the tests
started failing because the regex override doesn't propagate to the
tests.

In order to fix the failing tests, override the regex in the tests
as well.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ gemspec
 
 gem 'rgen'
 gem 'redcarpet'
-gem "yard", "~> 0.8.7"
+gem "yard", "~> 0.9.3"
 
 puppetversion = ENV['PUPPET_VERSION']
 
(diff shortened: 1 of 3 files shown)


HAIL9000:
Ping @hlindberg for thoughts on this. This may not be the right solution to the problem. I confirmed that when running strings (outside of the spec tests) with YARD 0.9.0 or greater the fix for [PDOC-25](https://tickets.puppetlabs.com/browse/PDOC-25) still works. Something is just going wrong and the override is not there in the spec tests. Happy to change this though if there is a better solution, this feels quite hacky.

HAIL9000:
Also I'm in favor of tying strings to a specific YARD version in the Gemfile and consciously making the decision to move to the latest version when it comes out rather than just defaulting to the latest, since a new YARD version has so much potential to break strings.

peterhuene:
I'm going to close this as we can't move straight to the latest Yard (which is now 0.9.5) as it breaks a few things in the generated HTML.  I'm almost done with an overhaul of puppet strings that cleans up the code base, polishes the command line / HTML output, and fixes a bunch of small bugs.  I'm going to close this PR in favor of that work, which should go up soon.

PR #94, opened 2016-08-12
(PDOC-75) Strings calls 'interpret_any' with the wrong number of arguments

eXprojects:
Puppet strings uses an API private method in puppet which has been updated to take two parameters instead of one. This means strings is calling it with the wrong number of variables in the defined type handler which is causing it to crash:

https://github.com/puppetlabs/puppetlabs-strings/blob/master/lib/puppet_x/puppetlabs/strings/yard/handlers/defined_type_handler.rb#L19

### DIFF
--- a/lib/puppet_x/puppetlabs/strings/yard/handlers/defined_type_handler.rb
+++ b/lib/puppet_x/puppetlabs/strings/yard/handlers/defined_type_handler.rb
@@ -16,7 +16,13 @@ class PuppetX::PuppetLabs::Strings::YARD::Handlers::DefinedTypeHandler < PuppetX
         param_type_info[pop_param.name] = Puppet::Pops::Types::TypeFactory.any()
       else
         begin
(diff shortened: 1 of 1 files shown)


peterhuene:
Hi @eXprojects.  Thank you very much for your contribution!

There is currently ongoing work to overhaul Puppet strings and move it to the latest version of Yard (see #98 which will hopefully be merged into master this week).  This work removes the use of the type parser in the defined type handler (and elsewhere), so unfortunately your fix is no longer needed.

The upcoming release of Puppet strings will not attempt to parse the type strings like it did before to verify that the docstring and the parameter types match; instead, it now always respects the Puppet version of the type and warns if you attempt to document the type in the docstring.  This is to prevent users from having to keep their documentation in sync with the code, which was problematic (especially for overloaded Puppet functions where parameter types may differ between overloads).

As such, I am closing this PR in favor of the refactoring work that will hopefully go in soon.  Please reopen this PR or message me if you have any concerns.  Thank you!

PR #97, opened 2016-09-01
(maint) Pin rubocop to enable spec support for ruby < 2.0

kylog:


kylog:
Whoops rubocop doesn't pass on strings atm .... Sigh. I'll update.

kylog:
Ping @peterhuene or @hlindberg.

peterhuene:
We should probably drop CI for < 2.0 soon, but I'm 👍 

My work includes moving us to the latest Yard and completely overhauling puppet-strings to produce a more polished output (wrt to command line and HTML output) and fixing lots of little bugs.  Should hopefully be ready to go by early next week at the latest.

kylog:
Maybe the polished output plus dropping old rubies would make for a 0.5.0 (or 1.0.0).

Actually we should probably ask some users of this gem about dropping the old rubies. FYI we had a similar conversation re puppet-lint in https://github.com/rodjek/puppet-lint/issues/503.

hlindberg:
👍  asking users; no point to support older versions if there are very few or noone using strings with those.

PR #103, opened 2016-09-28
(PDOC-80) Remove runtime dependency on puppet

whopper:
The runtime dependency on puppet presents problems when
puppet is not installed on the system as a gem, as is the
case in PE installations. Because the dependency will only be
satisfied by a gem, strings currently cannot be installed in
PE 3.8. This had to be worked around in newer PE versions with
a fake gemspec file.

Another issue is that installing strings with an older
version of puppet will force an upgrade to the newest version of
puppet via the gem.

This commit replaces the runtime dependency on puppet with a
requirement which will be present on the gem homepage on
rubygems.org. Requirements are meant to list external requirements
needed for the gem to work, which is correct for puppet as it can be
present via a different type of package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ In order to run strings you need to have the following software installed:
   * Puppet 3.7 or newer
   * The `yard` Ruby gem
 
+Note that a few extra steps are necessary to install puppet-strings with Puppet Enterprise 3.8.
+
 Installing the YARD Gem
 -----------------------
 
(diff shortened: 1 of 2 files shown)


whopper:
Tested the new instructions in PE 3.8, PE 2016.4 and FOSS with puppet 3.8 and 4.7 installed via the rubygem.

whopper:
The only slight downside to this is that you can now install the strings gem without puppet and it will just "work", but when you go to run it, it will error out with `puppet: command not found`. Really, though, all of our documentation says to use puppet to install the gem, so we don't think this is a problem. It doesn't make sense to install strings before or without puppet already present.

turbodog:
+1

hlindberg:
Everyone is happy - merging

PR #105, opened 2016-09-29
(PDOC-88) Add additional spec and acceptance tests

whopper:


spec/unit/puppet-strings/yard/handlers/ruby/provider_handler_spec.rb
```diff
@@ -27,6 +27,34 @@
     end
   end
 
+  describe 'parsing a provider with an invalid docstring assignment' do
+    let(:source) { <<-SOURCE
+Puppet::Type.type(:custom).provide :linux do
+  @doc = 123
```
er0ck: curiosity to me.. is this what you referenced for puppet 3x?  do the various types of docstrings have to be instance vars?

whopper: @er0ck I think you *can* reference the @doc instance var, but the only record of this being a common thing I could find was https://projects.puppetlabs.com/issues/11404.

I added the test to cover this bit of strings code: https://github.com/puppetlabs/puppet-strings/blob/master/lib/puppet-strings/yard/handlers/ruby/provider_handler.rb#L52-L55

spec/acceptance/running_strings_generate.rb
```diff
@@ -18,6 +18,10 @@ def read_file_on(host, filename)
     expect(read_file_on(master, '/root/doc/puppet_classes/test.html')).to include('Class: test')
   end
 
+  it 'should generate documentation for puppet functions' do
+    expect(read_file_on(master, '/root/doc/puppet_functions_puppet/strings_3A_3Aadd.html')).to include('Adds two integers together.')
```
er0ck: should this also test for the `@params` and `@return` and `@example`?

whopper: @er0ck yeah, it may as well. Updated to check for those.

whopper:
The acceptance test may not be working correctly - looking into it.

whopper:
Added two more tests for types and providers - all pass:
```
Generating module documentation using generate action
  should generate documentation for manifests
  should generate documentation for puppet functions
  should generate documentation for 3x functions
  should generate documentation for 4x functions
  should generate documentation for custom types
  should generate documentation for custom providers
```

PR #106, opened 2016-09-30
(PDOC-106) Add acceptance tests for --emit-json options

whopper:
These commits also add an acceptance helper module and add it to the LOAD_PATH.

### DIFF
--- a/spec/acceptance/emit_json_options.rb
+++ b/spec/acceptance/emit_json_options.rb
@@ -1,0 +1,41 @@
+require 'spec_helper_acceptance'
+require 'util'
+require 'json'
+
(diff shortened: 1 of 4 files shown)


whopper:
Tests passed locally:

```
Emitting JSON
  should emit JSON to stdout when using the --emit-json-stdout option
  should write JSON to a file when using the --emit-json option

...
Handing 'h478yqi6tww6pb6.delivery.puppetlabs.net' back to vmpooler for VM destruction
Spent 0.03 seconds cleaning up

Finished in 33.52 seconds (files took 18.08 seconds to load)
5 examples, 0 failures
```

PR #108, opened 2016-10-06
(PDOC-119) Add required ruby version to gemspec

whopper:
Strings is tested and supported for ruby 1.9.3 and above, so
the gemspec should reflect that.


whopper:
Note: you can't currently install strings with an older ruby anyway, since puppet itself requires 1.9.3. Now that we've removed the puppet runtime dependency, this requirement is more important.

whopper:
Actually @peterhuene would you mind clicking the button if this seems sane? Nobody else is around :)

PR #110, opened 2016-10-10
(PDOC-122) Properly parse `newfunction` calls with newlines

whopper:
When `newfunction` is separated from the Puppet::Parser::Functions module name by a
newline, YARD ignores the namespace and uses `newfunction` as the source of the
first statement.

Prior to this commit, strings didn't recognize this case, and 3.x functions written
in this way were not parsed as functions. This commit updates the ruby function handler
to identify and properly parse 3.x functions that include a newline between the
Puppet::Parser::Function namespace and the newfunction method call.

### DIFF
--- a/lib/puppet-strings/yard/handlers/ruby/function_handler.rb
+++ b/lib/puppet-strings/yard/handlers/ruby/function_handler.rb
@@ -23,12 +23,15 @@ class PuppetStrings::Yard::Handlers::Ruby::FunctionHandler < PuppetStrings::Yard
 
   process do
     # Only accept calls to Puppet::Functions (4.x) or Puppet::Parser::Functions (3.x)
+    # When `newfunction` is separated from the Puppet::Parser::Functions module name by a
+    # newline, YARD ignores the namespace and uses `newfunction` as the source of the
+    # first statement.
     return unless statement.count > 1
     module_name = statement[0].source
-    return unless module_name == 'Puppet::Functions' || module_name == 'Puppet::Parser::Functions'
+    return unless module_name == 'Puppet::Functions' || module_name == 'Puppet::Parser::Functions' || module_name == 'newfunction'
 
     # Create and register the function object
-    is_3x = module_name == 'Puppet::Parser::Functions'
+    is_3x = module_name == 'Puppet::Parser::Functions' || module_name == 'newfunction'
     object = PuppetStrings::Yard::CodeObjects::Function.new(
       get_name,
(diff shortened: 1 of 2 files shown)


whopper:
/cc @peterhuene (I'm only picking on you because Jesse is out on vacation :))

Note also that this weird newline business is actually how we still document writing 3.x functions: https://docs.puppet.com/guides/custom_functions.html

lib/puppet-strings/yard/handlers/ruby/function_handler.rb
```diff
@@ -23,12 +23,15 @@ class PuppetStrings::Yard::Handlers::Ruby::FunctionHandler < PuppetStrings::Yard
 
   process do
     # Only accept calls to Puppet::Functions (4.x) or Puppet::Parser::Functions (3.x)
+    # When `newfunction` is separated from the Puppet::Parser::Functions module name by a
+    # newline, YARD ignores the namespace and uses `newfunction` as the source of the
+    # first statement.
     return unless statement.count > 1
     module_name = statement[0].source
-    return unless module_name == 'Puppet::Functions' || module_name == 'Puppet::Parser::Functions'
+    return unless module_name == 'Puppet::Functions' || module_name == 'Puppet::Parser::Functions' || module_name == 'newfunction'
```
whopper: Naming is hard. If `statement[0]` is `newfunction`, it's not really a module name. Worth renaming?

peterhuene: I think this is fine.

peterhuene:
👍  although Ruby developers that open a module to invoke a function on it should be shot.

peterhuene:
We should open a ticket to update the documentation to rid ourselves of that bad style.

spec/unit/puppet-strings/yard/handlers/ruby/function_handler_spec.rb
```diff
@@ -70,6 +70,50 @@
...
+        expect(tags.size).to eq(1)
+        expect(tags[0].name).to be_nil
+        expect(tags[0].text).to eq('Returns nothing.')
+        expect(tags[0].types).to eq(['Undef'])
+        tags = object.docstring.tags(:api)
+        expect(tags.size).to eq(1)
+        expect(tags[0].text).to eq('public')
+      end
+    end
+
```
er0ck: yay.  looks good to me.  no need for acceptance

PR #111, opened 2016-10-11
(PDOC-95) Properly group and display multiple provider `defaultfor`s

whopper:
Prior to this commit, strings did not properly handle providers
which had multiple related `defaultfor`s. In code, these are written
as comma-separated constraints. This commit updates strings' puppet
provider handler, as well as the template which generates HTML for
`defaultfor` statements.

Note that it was necessary to make a breaking change to the JSON
schema to accomodate multiple AND'ed defaults. Previously, provider
defaults were contained in a single key-value map. Now, they are
contained in an array of key-value maps, which allows multiple
constraints to be associated with each other.

### DIFF
--- a/JSON.md
+++ b/JSON.md
@@ -3,7 +3,7 @@ Puppet Strings JSON Data
 
 Puppet Strings has two flags to the `generate` action that can be used to emit JSON data:
 
-* `--emit-json <file>`: Emits the JSON data to the given file. 
+* `--emit-json <file>`: Emits the JSON data to the given file.
 * `--emit-json-stdout`: Emits the JSON data to STDOUT.
 
 Document Schema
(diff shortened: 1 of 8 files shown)


whopper:
Here's a before and after for the systemd provider, which has the following defaults:

```
  defaultfor :osfamily => [:archlinux]
  defaultfor :osfamily => :redhat, :operatingsystemmajrelease => "7"
  defaultfor :osfamily => :redhat, :operatingsystem => :fedora
  defaultfor :osfamily => :suse
  defaultfor :operatingsystem => :debian, :operatingsystemmajrelease => "8"
  defaultfor :operatingsystem => :ubuntu, :operatingsystemmajrelease => ["15.04","15.10","16.04","16.10"]
  defaultfor :operatingsystem => :cumuluslinux, :operatingsystemmajrelease => ["3"]
```

**Before**:

screen shot 2016-10-11 at 12 57 22 pm

**After**:
screen shot 2016-10-11 at 12 57 37 pm

peterhuene:
👍

peterhuene:
Question though: is `defaultfor osfamily: 'RedHat', osfamily: 'Gentoo'` valid?  That is to say, should it be an array of key value pairs rather than a map?

whopper:
@peterhuene `defaultfor osfamily: 'RedHat', osfamily: 'Gentoo'` would be a conflicting `defaultfor` and would make the provider not the default of anything.

whopper:
@peterhuene updated to use list of lists and verified the case of defaultfor osfamily: 'RedHat', osfamily: 'Gentoo' is displayed correctly (and that everything else still works)

whopper:
Ah, need to fix the schema.

PR #114, opened 2016-10-12
(PDOC-128) Add explanation of @example tag usage in README

whopper:


nfagerlund:
👍 Seems clear enough to me!

PR #115, opened 2016-10-13
(PDOC-120) Update gh_pages task to allow publishing of _index.html

whopper:
Previously, due to Jekyll's default rule of not allowing files beginning
with an '_' to be published, the gh_pages rake task led to broken links.
This commit adds a step to the task which adds a _config.yml file to the
`doc` directory of the gh-pages branch which explicitly allows the file
to be published.


whopper:
And it even works!

screen shot 2016-10-13 at 12 16 52 pm

PR #118, opened 2016-10-26
(PDOC-126) Remove `%Q` ruby quotation syntax from parsed strings

whopper:
Previously, YARD would parse ruby strings which used %Q notation
and return the syntax literally. This commit adds a new Util
module with a `scrub_string` method to check for such a string
and remove the errant %Q and %q's.

### DIFF
--- a/lib/puppet-strings/yard/util.rb
+++ b/lib/puppet-strings/yard/util.rb
@@ -1,0 +1,17 @@
+require 'puppet/util'
+
+# The module for various puppet-strings utility helpers.
+module PuppetStrings::Yard::Util
+  # Trims indentation from trailing whitespace and removes ruby literal quotation
+  # syntax `%Q{}` and `%{q}` from parsed strings.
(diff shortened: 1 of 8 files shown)


lib/puppet-strings/yard/util.rb
```diff
@@ -0,0 +1,17 @@
+require 'puppet/util'
+
+# The module for various puppet-strings utility helpers.
+module PuppetStrings::Yard::Util
+  # Trims indentation from trailing whitespace and removes ruby literal quotation
+  # syntax `%Q{}` and `%{q}` from parsed strings.
+  # @param [String] str The string to scrub.
+  # @return [String] A scrubbed string.
+  def self.scrub_string(str)
+    match = str.match(/^%[Qq]{(.*)}/m)
```
scotje: maybe add the end of string anchor to this regex

whopper: @scotje fixed!

PR #121, opened 2016-11-03
(PDOC-121) Include tag or SHA in gh_pages task commit

whopper:
This commit updates the gh_pages task to reference the SHA
of the branch being documented in the documentation commit.

In addition, the description of each task except `update`
has been removed, as they are not useful on their own and
should be considered private. This is necessary in order
to hide the tasks from the output of `rake -T`.

### DIFF
--- a/lib/puppet-strings/tasks/gh_pages.rb
+++ b/lib/puppet-strings/tasks/gh_pages.rb
@@ -2,7 +2,6 @@ require 'puppet-strings/tasks'
 
 namespace :strings do
   namespace :gh_pages do
-    desc 'Checkout the gh-pages branch for doc generation.'
     task :checkout do
       if Dir.exist?('doc')
         fail "The 'doc' directory (#{File.expand_path('doc')}) is not a Git repository! Remove it and run the Rake task again." unless Dir.exist?('doc/.git')
@@ -24,7 +23,6 @@ namespace :strings do
       end
     end
 
-    desc 'Add Jekyll _config.yml file to allow publishing of _index.html.'
     task :configure do
       unless File.exist?(File.join('doc', '_config.yml'))
         Dir.chdir('doc') do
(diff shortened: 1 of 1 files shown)


lib/puppet-strings/tasks/gh_pages.rb
```diff
@@ -37,13 +44,14 @@
     task :push do
       Dir.chdir('doc') do
         system 'git add .'
-        system "git commit -m '[strings] Generated Documentation Update'"
+        system "git commit -m '[strings] Generated Documentation Update at Revision #{@git_sha}'"
```
whopper: ah, I hadn't thought of that. Good idea :D

whopper: @scotje updated

lib/puppet-strings/tasks/gh_pages.rb
```diff
@@ -34,16 +41,17 @@
...
         system 'git push origin gh-pages -f'
       end
     end
 
     desc 'Run checkout, generate, and push tasks.'
     task :update => [
+      :get_sha,
```
scotje: don't know that it hurts anything but this is probably unnecessary now

whopper: Indeed, updated.

whopper:
@scotje updated again

lib/puppet-strings/tasks/gh_pages.rb
```diff
@@ -33,16 +31,19 @@
...
       Dir.chdir('doc') do
         system 'git add .'
-        system "git commit -m '[strings] Generated Documentation Update'"
+        system "git commit -m '[strings] Generated Documentation Update at Revision #{git_sha}'"
         system 'git push origin gh-pages -f'
       end
     end
 
-    desc 'Run checkout, generate, and push tasks.'
+    desc 'Run checkout, generate, configure, and push tasks.'
```
scotje: One last thing, we should probably make this description just say "update docs and push to github" or something like that, since we aren't advertising the underlying tasks anymore.

PR #119, opened 2016-11-01
(PDOC-125) Display all puppet function signatures in top-level signature key

whopper:

PR #122, opened 2016-11-04
(PDOC-129) Include tags in overload objects when serialized as JSON

whopper:
Previously, overload objects were not displaying their tags when
they had no docstring text. This was due to an issue in the overload
`to_hash` method, which prevented the tags from being serialized when
the dispatch had no top-level text. This commit updates that logic
so that the tags will always be included in the hash if they exist.

### DIFF
--- a/lib/puppet-strings/yard/tags/overload_tag.rb
+++ b/lib/puppet-strings/yard/tags/overload_tag.rb
@@ -99,7 +99,7 @@ class PuppetStrings::Yard::Tags::OverloadTag < YARD::Tags::Tag
     hash[:tag_name] = tag_name
     hash[:text] = text if text
     hash[:signature] = signature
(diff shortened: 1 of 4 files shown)


whopper:
This should probably be merged after https://github.com/puppetlabs/puppet-strings/pull/119 to prevent a bigger rebasing headache. I'll have to rebase this once that's merged, though.

whopper:
@scotje indeed, it's a bit dumb. Actually, we may be able to use the `Docstring.blank?` method: http://www.rubydoc.info/gems/yard/YARD/Docstring#blank%3F-instance_method

scotje:
Yeah, it looks like `blank?` does what we want in this case?

whopper:
@scotje updated to use `blank?` and verified it works as expected.

scotje:
Hooray!

hlindberg:
LGTM

whopper:
@scotje rebased!

PR #123, opened 2016-11-07
(PDOC-93) Ensure search bar doesn't overlap item list in nav bar

whopper:


whopper:
Before this change:

screen shot 2016-11-07 at 2 26 18 pm


With this change:

screen shot 2016-11-07 at 2 25 20 pm

screen shot 2016-11-07 at 2 25 04 pm

PR #125, opened 2016-11-10
(PDOC-135) Detect `return_type` calls in 4.x function dispatches

whopper:
Previously, Strings ignored calls to `return_type` in Puppet 4.x API
function dispatches, preventing the return types of overloads from
being automatically determined. This commit adds a check for a node
with a `return_type` call and handles it properly.

### DIFF
--- a/lib/puppet-strings/yard/handlers/ruby/function_handler.rb
+++ b/lib/puppet-strings/yard/handlers/ruby/function_handler.rb
@@ -15,6 +15,7 @@ class PuppetStrings::Yard::Handlers::Ruby::FunctionHandler < PuppetStrings::Yard
     block_param
     required_block_param
     optional_block_param
+    return_type
   ).freeze
 
   namespace_only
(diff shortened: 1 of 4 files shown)


lib/puppet-strings/yard/handlers/ruby/function_handler.rb
```diff
@@ -133,36 +134,40 @@ def add_overload_tag(object, node)
       method_name = child.method_name.source
       next unless DISPATCH_METHOD_NAMES.include?(method_name)
 
-      # Check for block
-      if method_name.include?('block')
-        if block
-          log.warn "A duplicate block parameter was found for Puppet function '#{object.name}' at #{child.file}:#{child.line}."
+      if method_name == 'return_type'
+        overload_tag.tag(:return).types = [node_as_string(child.parameters[0])]
+      else
```
hlindberg: could just do a `next` to get out of this conditional. Then there is no need to nest the rest of the logic under this conditional's `else` (right now there is a construct `else` `if`` which should really be an `elseif` - but since all other cases use `next` this case ('return_type') could use that as well as it is easier to read without the nesting.

whopper: @hlindberg updated.

PR #126, opened 2016-11-11
(PDOC-136) Detect return type syntax in Puppet Language functions

whopper:
Previously, Strings ignored the return type in Puppet 3.x functions
that used the following syntax:

function foo() >> String {}

This commit updates the FunctionStatement class to use the return
type from such a statement if it exists.

### DIFF
--- a/lib/puppet-strings/yard/handlers/puppet/function_handler.rb
+++ b/lib/puppet-strings/yard/handlers/puppet/function_handler.rb
@@ -20,7 +20,7 @@ class PuppetStrings::Yard::Handlers::Puppet::FunctionHandler < PuppetStrings::Ya
     set_parameter_types(object)
 
     # Add a return tag
-    add_return_tag(object)
+    add_return_tag(object, statement.type)
 
     # Set the parameters on the object
     object.parameters = statement.parameters.map { |p| [p.name, p.value] }
@@ -30,13 +30,18 @@ class PuppetStrings::Yard::Handlers::Puppet::FunctionHandler < PuppetStrings::Ya
   end
 
   private
-  def add_return_tag(object)
+  def add_return_tag(object, type=nil)
     tag = object.tag(:return)
     if tag
-      tag.types = ['Any'] unless tag.types
+      if (type && tag.types) && (type != tag.types)
+        log.warn "Documented return type does not match return type in function definition near #{statement.file}:#{statement.line}."
+      end
+
+      tag.types = type ? [type] : tag.types || ['Any']
       return
     end
     log.warn "Missing @return tag near #{statement.file}:#{statement.line}."
-    object.add_tag YARD::Tags::Tag.new(:return, '', 'Any')
+    type = type || 'Any'
+    object.add_tag YARD::Tags::Tag.new(:return, '', type)
   end
 end
--- a/lib/puppet-strings/yard/parsers/puppet/statement.rb
+++ b/lib/puppet-strings/yard/parsers/puppet/statement.rb
@@ -134,6 +134,7 @@ module PuppetStrings::Yard::Parsers::Puppet
   # Implements the Puppet function statement.
   class FunctionStatement < ParameterizedStatement
     attr_reader :name
+    attr_reader :type
 
     # Initializes the Puppet function statement.
     # @param [Puppet::Pops::Model::FunctionDefinition] object The model object for the function statement.
@@ -141,6 +142,13 @@ module PuppetStrings::Yard::Parsers::Puppet
     def initialize(object, file)
       super(object, file)
       @name = object.name
+      if object.respond_to? :return_type
+        type = object.return_type
+        if type
+          adapter = ::Puppet::Pops::Adapters::SourcePosAdapter.adapt(type)
+          @type = adapter.extract_text.gsub('>> ', '')
+        end
+      end
     end
   end
 end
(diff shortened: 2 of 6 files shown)


spec/unit/puppet-strings/yard/handlers/puppet/function_handler_spec.rb
```diff
@@ -68,6 +68,11 @@
       expect(tags[2].name).to eq('param3')
       expect(tags[2].text).to eq('Third param.')
       expect(tags[2].types).to eq(['String'])
+      tags = object.docstring.tags(:return)
+      expect(tags.size).to eq(1)
+      expect(tags[0].tag_name).to eq('return')
+      expect(tags[0].text).to eq('Returns nothing.')
+      expect(tags[0].types).to eq(['Undef'])
```
scotje: Future TODO item: de-compose all the different expectations in this test.

lib/puppet-strings/yard/parsers/puppet/statement.rb
```diff
@@ -134,13 +134,15 @@ def initialize(object, file)
...
     attr_reader :name
+    attr_reader :type
 
     # Initializes the Puppet function statement.
     # @param [Puppet::Pops::Model::FunctionDefinition] object The model object for the function statement.
     # @param [String] file The file containing the statement.
     def initialize(object, file)
       super(object, file)
       @name = object.name
+      @type = object.return_type.value.capitalize if object.return_type
```
hlindberg: Note that `object_type` was introduced quite recently and this will fail if using puppet < 4.7 or so (don't remember exactly when return_type was added). Code should protect against that if Strings can use a puppet older than when it was introduced.

The return type can be a static expression, it is not enough to assume it is a string like value. Needs to do whatever is done for parameter types (have not looked, I am assuming it does it correctly to handle things like Integer[1,10] and other parameterized types).

spec/unit/puppet-strings/yard/handlers/puppet/function_handler_spec.rb
```diff
@@ -166,4 +171,70 @@
...
+      expect(subject.size).to eq(1)
+      object = subject.first
+      expect(object).to be_a(PuppetStrings::Yard::CodeObjects::Function)
+      tags = object.docstring.tags(:return)
+      expect(tags.size).to eq(1)
+      expect(tags[0].tag_name).to eq('return')
+      expect(tags[0].text).to eq('this is a lie.')
+      expect(tags[0].types).to eq(['String'])
+    end
+  end
```
scotje: Seems like this case should also emit a warning if possible.

scotje: (Maybe it does and we just need to add the expectation for it to the test?)

spec/unit/puppet-strings/json_spec.rb
```diff
@@ -27,8 +27,8 @@ class klass(Integer $param1, $param2, String $param3 = hi) inherits foo::bar {
 # @param param1 First param.
 # @param param2 Second param.
 # @param param3 Third param.
-# @return [Undef] Returns nothing.
-function func(Integer $param1, $param2, String $param3 = hi) {
+# @return Returns nothing.
+function func(Integer $param1, $param2, String $param3 = hi) >> Undef {
```
hlindberg: A test with a complex return type is needed - like a Struct

spec/unit/puppet-strings/yard/handlers/puppet/function_handler_spec.rb
whopper: Agreed. That and we should make the big JSON test more sane.

spec/unit/puppet-strings/yard/handlers/puppet/function_handler_spec.rb
whopper: I had the same thought. Adding a warning.

lib/puppet-strings/yard/handlers/puppet/function_handler.rb
```diff
@@ -30,13 +30,18 @@ class PuppetStrings::Yard::Handlers::Puppet::FunctionHandler < PuppetStrings::Ya
   end
 
   private
-  def add_return_tag(object)
+  def add_return_tag(object, type=nil)
     tag = object.tag(:return)
     if tag
-      tag.types = ['Any'] unless tag.types
+      if (type && tag.types) && (type != tag.types)
+        log.warn "Documented return type does not match return type in function definition near #{statement.file}:#{statement.line}."
```
whopper: @scotje added this warning! ^

lib/puppet-strings/yard/parsers/puppet/statement.rb
whopper: @hlindberg added a check to make sure the method exists. Also updated to use `Puppet::Pops::Adapters::SourcePosAdapter.adapt` to extract the type.

spec/unit/puppet-strings/json_spec.rb
whopper: Added a complex example to the function handler and parser spec, but left the json test alone. I think the JSON test is starting to go beyond the scope of what it should do, which is just ensure JSON output works and matches our schema.

whopper:
@scotje @hlindberg fixed up and travis is green now :D

PR #128, opened 2016-11-23
(PDOC-133) revise+edit strings readme

jbondpdx:
@whopper , here's my revision on the Strings README!

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,6 @@ A Puppet command built on [YARD](http://yardoc.org/).
 
 Puppet Strings generates HTML documentation for Puppet extensions written in Puppet and Ruby.
 
-This tool will eventually place the existing `puppet doc` command once feature parity has been achieved.
-
 |                |                                                                 |
 | -------------- |---------------------------------------------------------------- |
 | *Code*         | [GitHub][repo]                                                  |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,123 +21,155 @@ This tool will eventually place the existing `puppet doc` command once feature p
...
+
 ```
 $ puppet resource package yard provider=gem
 ```
 
 For Puppet Enterprise 3.8:
+
 ```
 GEM_HOME=/opt/puppet/lib/ruby/gems/1.9.1 puppet resource package yard provider=gem
 ```
```
er0ck: does this work in windows?  should we provide instructions there?

README.md
```diff
@@ -23,123 +21,155 @@ This tool will eventually place the existing `puppet doc` command once feature p
...
 
 ```
-$ puppet strings generate --emit-json documentation.json
+$ puppet strings server
 ```
 
-It can also print the JSON to stdout:
+### Output documents in JSON
+
+Strings can produce documentation in JSON and then either generate a `.json` file or print JSON to stdout. JSON. This can be useful for handling or displaying the data with your own custom applications.
```
whopper: I think that "JSON." in the middle of the sentence is a typo.

README.md
```diff
@@ -199,13 +245,8 @@ DESC
...
 ```
-
-Puppet Strings documents this way to preserve backwards compatibility with `puppet doc` and existing resource types.
-
-***Note: Puppet Strings does not evaluate your Ruby code, so only certain static expressions are supported.***
-
-To document parameters and properties that are dynamically created, use the `#@!puppet.type.param` and `#@!puppet.type.property`
-directives before the `newtype` call:
+ 
+If your resource type includes dynamically create parameters and properties, you must also use the `#@!puppet.type.param` and `#@!puppet.type.property` directives **before** the `newtype` call. This is necessary because Strings does not evaluate Ruby code, so it cannot detect dynamic attributes.
```
whopper: Typo: "If your resource type includes dynamically create parameters and properties"

README.md
```diff
@@ -227,39 +266,17 @@ Puppet::Type.type(:example).provide :platform do
 end
 ```
 
-Puppet Strings documents this way to preserve backwards compatibility with `puppet doc` and existing resource types.
-
-***Note: Puppet Strings does not evaluate your Ruby code, so only certain static expressions are supported.***
-
-### Functions
+All provider method calls, including confine, defaultfor, and commands, are automatically parsed and documented by Strings. The `desc` method is used to generate the docstring, and can include tags such as @example if written as a heredoc.
```
whopper: Should 'confine', 'defaultfor', 'commands' and @example be in backticks here?

README.md
```diff
@@ -227,39 +266,17 @@ Puppet::Type.type(:example).provide :platform do
...
 
-Puppet Strings supports three different ways of defining a function in Puppet: with the Puppet 3.x API, Puppet 4.X API,
-and in the Puppet language itself.
+**Note**: Puppet Strings can not evaluate your Ruby code, so only certain static expressions are supported.
 
-#### Puppet 3.x Functions
+### Documenting functions
 
-To document a function in the Puppet 3.x API, use the `doc` option to `newfunction`:
+Puppet Strings supports defining functions with the Puppet 4 API, the Puppet 3 API, or in the Puppet language itself.
```
er0ck: should this say "supports the documenting of defined functions with"?

README.md
```diff
@@ -351,103 +382,124 @@ The string following the `@example` tag is an optional title which is displayed
...
+You can use Puppet Strings rake tasks to generate documentation with additional options or to make your generated docs available on [GitHub Pages](https://pages.github.com/).
+
+The `strings:generate` and `strings:gh_pages:update` tasks are available in `puppet-strings/tasks`.
+
+First, update your Gemfile and your Rakefile.:
+
+1. Add the following to your Gemfile to use `puppet-strings`:
+
+   ```ruby
 gem 'puppet-strings', :git => 'https://github.com/puppetlabs/puppet-strings.git'
```
er0ck: why does using the rake tasks require installing Strings from git, rather than say rubygems.org?

whopper:
@jbondpdx would you mind also slightly modifying the commit message to have (PDOC-133) rather than 'PDOC-133:'? It doesn't *really* matter but I figure we may as well be consistent :)

README.md
```diff
@@ -351,103 +382,124 @@ The string following the `@example` tag is an optional title which is displayed
...
-```
+   ```
 
-In your `Rakefile`, add the following to use the `puppet-strings` tasks:
+2. Add the following to your `Rakefile` to use the `puppet-strings` tasks:
 
-```ruby
+   ```ruby
 require 'puppet-strings/tasks'
```
er0ck: i haven't looked in here yet, but does this implicitly define the tasks below?

README.md
whopper: It doesn't - this can just be `gem 'puppet-strings'`

README.md
whopper: Yep, this does this:

```
    require 'puppet-strings/tasks/generate.rb'
    require 'puppet-strings/tasks/gh_pages.rb'
```

README.md
jbondpdx: Oh, indeed it is. Will fix.

README.md
jbondpdx: I have no idea! And @er0ck , your question applies to the whole doc, right? @whopper ?

README.md
jbondpdx: good catch

README.md
jbondpdx: I don't know the answer to that. @whopper ?

README.md
jbondpdx: @er0ck I'm sorry, I'm not sure what you mean exactly. Can you say more?

README.md
er0ck: yes, i suppose it applies to the whole doc.  It appears whopper wants to handle this later, which is fine with me.

README.md
er0ck: i think the `:git` use just confuses things, if not required.

README.md
er0ck: i think Will answered my question.  We may want a statement here that says something like "using this require will automatically create the Rake tasks below"

README.md
whopper: @jbondpdx yeah, I'd remove the `:git` part. It works with the gem from rubygems just fine.

README.md
whopper: Strings should work fine on Windows, but we haven't done any testing around it. I think it warrants its own investigation / ticket.

README.md
jbondpdx: Cool. I imagine it's some remnant of bits and pieces being added to this readme over a long period of time.

README.md
jbondpdx: Ah, that's very helpful, perfect!

README.md
er0ck: i'll create a ticket to expand the acceptance matrix

whopper:
👍 from me. Any further comments, @er0ck?

er0ck:
👍 :shipit:

PR #130, opened 2016-12-02
(PDOC-143) Add installation instructions for PE 3.8 in Windows

whopper:


jbondpdx:
Yay! This looks good to me, +1

whopper:
Whoa, that's a weird rubocop failure. Looks like it had an update or something and now I need to rearrange some of the gems. I'll add it as a second commit.

PR #131, opened 2016-12-20
(maint) Fix @param type examples in README

HAIL9000:
Correct the examples of using a parameter type in the @param tag
so that they match how the YARD documentation suggests the tag
should be used.

http://www.rubydoc.info/gems/yard/file/docs/Tags.md#param


scotje:
Thanks @HAIL9000 !

PR #132, opened 2017-01-10
(PDOC-155) Allow type documentation in Puppet 4 code

whopper:
Because Puppet 4 is typed, parameter type information can
be automatically determined without any explicit documentation
in @param tags. However, users may desire to do so anyway for
consistency. This commit allows @param tags to include a [type]
in Puppet 4 code. Strings will emit a warning if the documented
type does not match the actual type. In such an event, the incorrect
documented type will be ignored in favor of the real one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,14 +210,14 @@ The Strings elements appearing in the above comment block are:
 * The `include` statement, which adds the usage example code.
 * Two `@param` tags, with the name of the parameter first, followed by a string describing the parameter's purpose.
 
-Puppet 4 is a typed language, so Puppet Strings automatically documents the parameter types from code. With Puppet 3, however, include the parameter type with the `@param` tag:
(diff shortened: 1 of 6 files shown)


jbondpdx:
<3

README.md
```diff
@@ -210,15 +210,15 @@ The Strings elements appearing in the above comment block are:
 * The `include` statement, which adds the usage example code.
 * Two `@param` tags, with the name of the parameter first, followed by a string describing the parameter's purpose.
 
-Puppet 4 is a typed language, so Puppet Strings automatically documents the parameter types from code. With Puppet 3, however, include the parameter type with the `@param` tag:
+Puppet 4 is a typed language, so Puppet Strings automatically documents the parameter types from code. You may optionally include a parameter type in the `@param` tag. Strings will emit a warning should the documented type differ from the actual type.
+
```
HAIL9000: Maybe worth mentioning that in this case it will default to what it gets from the code rather than the comment? Unless that's already mentioned elsewhere in the README

whopper: @HAIL9000 updated

PR #135, opened 2017-02-07
(pdoc-144) acceptance: fix windows pre suite

er0ck:
* windows install requires type=aio so gem_command can find the right gem exe
* this change also allows easier command line use
  * by using a sensible default for the acceptance key

* install the correct CAs in the correct places so windows can install gems
* use a gem_command abstraction to find the correct puppet gem exe

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -21,12 +21,17 @@ end
 task :acceptance do
   require 'beaker-hostgenerator'
 
+  install_type = 'aio'
   target = ENV['platform']
+  unless target =~ /type=/
+    puts "INFO: adding 'type=#{install_type}' to host config"
+    target += "{type=#{install_type}}"
+  end
+
   if ! target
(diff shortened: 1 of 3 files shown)


whopper:
@er0ck I actually have no memory of how to run these tests :'D What else do I need besides`be rake acceptance` again?

whopper:
Nevermind, just needed `platform=<blah>`

er0ck:
yep.  no longer need to specify the acceptance key, if it's in the right place.

whopper:
@er0ck how do I give beaker a master when running with a windows agent?

```
     #   There should be one host with master defined!
```

whopper:
👍

er0ck:
FYI: future us:
the tests will still not pass on windows.  we can come back and fix that small detail later

PR #136, opened 2017-03-02
Fix args handling for rake strings::generate

hashar:
In the rake task strings::generate, the arguments json and yard_args
are not applied properly. The logic uses args.key?, however args is not
a Hash but a  Rake::TaskArguments (and Enumerable). There is no key?
method.

Use Rake::TaskArguments.has_key? instead.

### DIFF
--- a/lib/puppet-strings/tasks/generate.rb
+++ b/lib/puppet-strings/tasks/generate.rb
@@ -14,8 +14,11 @@ namespace :strings do
       markup: args[:markup] || 'markdown',
     }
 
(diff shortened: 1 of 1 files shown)


scotje:
Can you rebase this on master? That should fix the Travis/Rubocop error

hashar:
I think I have managed to rebase it

hashar:
Looks like Rubocop get confused. I have amended the change to selectively ignore the rule `Style/PreferredHashMethods` around those two lines.

ScottGarman:
:+1:

PR #137, opened 2017-03-11
(PDOC-160) Remove the timestamp from output

GeoffWilliams:
Prevent mass-changes to files by removing the date in the footer of each page.
Since the generation date isn't particularly useful and isn't tied to either
a git revision or a released version, this could be an easy way to address this
bug.  The files will of-course be regenerated but if there are no changes then
the content will be identical and git will not notice.

Solution courtesy of http://stackoverflow.com/a/11444144


GeoffWilliams:
Failed tests not related to this patch

scotje:
A rebase on master at this point should resolve the Travis/Rubocop errors.

Is it common to check the generated documentation into version control? Would a .gitignore entry for the generated documentation be more appropriate?

GeoffWilliams:
Thanks @scotje I'll give that a go next time I get a moment.

I personally think checking the generated docs into git is a good idea because:
* You don't need to functioning puppet-strings toolchain to read the docs
* You can track documentation changes - including those made by new features to puppet-strings if such things interest you
* Most importantly - you can upload the generated docs in your module and use githubs rawgit.com mirror to view them.  Now you just need to drop a link like https://rawgit.com/GeoffWilliams/puppet-motd/master/doc/index.html into your README.md and the documentation is done
* Best of all, you can now click-thru from the README page to the api-style docs from your module on the puppet forge.  This can be on any platform including mobile and means you can easily lookup parameter names, default values, datatypes and examples all without even opening a computer

GeoffWilliams:
@scotje - that worked.  Rebase fixed the CI fails

scotje:
OK, I don't think it's a big deal as far as this change goes. We're still hoping to eventually auto-generate and host the strings docs on the Forge as well.

PR #138, opened 2017-03-15
(PDOC-161) Add `summary` tag for short descriptions

whopper:
This tag is primarily meant to be applied to Puppet classes,
but it is also supported by every other construct that can be
documented with strings.

### DIFF
--- a/lib/puppet-strings/yard/handlers/helpers.rb
+++ b/lib/puppet-strings/yard/handlers/helpers.rb
@@ -1,0 +1,8 @@
+module PuppetStrings::Yard::Handlers::Helpers
+  # Logs a warning if a summary tag has more than 140 characters
+  def self.validate_summary_tag(object)
+    if object.has_tag?(:summary) && object.tag(:summary).text.length > 140
+      log.warn "The length of the summary for #{object.type} '#{object.name}' exceeds the recommended limit of 140 characters."
+    end
+  end
+end
--- a/lib/puppet-strings/yard/handlers/puppet/function_handler.rb
+++ b/lib/puppet-strings/yard/handlers/puppet/function_handler.rb
@@ -1,3 +1,4 @@
+require 'puppet-strings/yard/handlers/helpers'
 require 'puppet-strings/yard/handlers/puppet/base'
 require 'puppet-strings/yard/parsers'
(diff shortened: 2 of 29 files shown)


lib/puppet-strings/yard/handlers/puppet/function_handler.rb
```diff
@@ -27,6 +27,11 @@ class PuppetStrings::Yard::Handlers::Puppet::FunctionHandler < PuppetStrings::Ya
 
     # Mark the class as public if it doesn't already have an api tag
     object.add_tag YARD::Tags::Tag.new(:api, 'public') unless object.has_tag? :api
+
+    # Warn if a summary longer than 140 characters was provided
+    if object.has_tag?(:summary) && object.tag(:summary).text.length > 140
+      log.warn "The length of the summary for function '#{object.name}' exceeds the recommended limit of 140 characters."
```
scotje: Should we distinguish between Puppet functions and Ruby functions in the warning?

whopper:
@scotje I toyed with the idea of a new class on top of the Ruby and Puppet handlers to deal with that error message but I thought it ended up being more gross. I figured I'd just follow the convention that already exists for adding a public api tag in each of the handlers and do it in each. I could look into simplifying it again if you want.

scotje:
Is there somewhere we could put like a class method just so we have the logic and message all in one place, even if we still have to invoke it from each handler? (With an argument for the context it was invoked in so you can construct a specific message still.)

whopper:
@scotje sure, that seems reasonable. I'll look into it.

jpadams:
I built the gem from @whopper 's branch and it worked just as expected. +1 to merge.

lib/puppet-strings/yard/handlers/helpers.rb
```diff
@@ -0,0 +1,8 @@
+module PuppetStrings::Yard::Handlers::Helpers
+  # Logs a warning if a summary tag has more than 140 characters
+  def self.validate_summary_tag(object)
+    if object.has_tag?(:summary) && object.tag(:summary).text.length > 140
```
whopper: @scotje I stuck this in this Helper module for the Handlers. Does that seem reasonable?

PR #143, opened 2017-07-26
Use the supported ruby version for each Puppet version.

ghoneycutt:
This patch moves from using a matrix of ruby and puppet version and then
excluding the incompatibilities to including supported versions. This
makes it much easier to see what is supported while also ensuring the
correct matrix of tests.

https://docs.puppet.com/puppet/4.10/about_agent.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,39 +8,40 @@ bundler_args: --without development acceptance
 notifications:
   email: false
 
-rvm:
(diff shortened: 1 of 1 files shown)


Magisus:
This seems very similar to #144. Which is the right one?

Magisus:
This is superseded by #144.

PR #144, opened 2017-07-26
Drop support for old releases

ghoneycutt:


Magisus:
Should this also include entries for Puppet 5/Ruby 2.4.1?

Magisus:
Does this supersede #143 ?

ghoneycutt:
Hi @Magisus 

This does supercede PR #143.

It does not include Puppet 5 / Ruby 2.4.1 as that was not already supported.

It does not include ruby 2.3.3 as that was never a supported Ruby version by Puppet.

Magisus:
Alright. In that case this looks fine to me. Shall I close the other one and merge this one?

PR #146, opened 2017-10-04
(QENG-5344) Add beaker-abs for CI.next-i-fication

smcelmurry:
This is needed in order to work with always be scheduling on CI.Next jenkins instances.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ bundler_args: --without development acceptance
 notifications:
   email: false
 
+before_install:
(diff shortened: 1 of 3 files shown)


sbeaulie:
Any idea why travis fails?

smcelmurry:
Ping @scotje for review

scotje:
Looks like we just need a newer version of bundler under 1.9.3. I pushed a commit that should hopefully do this in Travis.

PR #151, opened 2018-01-19
(PDOC-196) Correcting misinformation about supported YARD tags

jbondpdx:
The readme says we support all YARD tags. Turns out, we don't!

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,15 +403,14 @@ function example(string $name) {
 
 The most commonly used tags for Strings are:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -403,15 +403,14 @@ function example(string $name) {
 
 The most commonly used tags for Strings are:
```
eputnam: Maybe get rid of this line? Take out the grey area, these are the tags you can use. What do you think?

jbondpdx: Seems legit to me! On it.

PR #152, opened 2018-01-26
Enable code coverage tracking

DavidS:


eputnam:
why do we want/need code coverage tracking?

DavidS:
It provides guidance into what parts are lacking in test coverage. I'm enabling it on all repos I'm touching.

PR #156, opened 2018-02-07
(PDOC-184) generate markdown

eputnam:
This change does a few things:
1. Fixes up new api handler to return the stuff we want
2. Adds all the logic to parse YARD registries into markdown
3. Adds templates for markdown
4. Changes Face cli to use a --format option that can be used for either
markdown or json


lib/puppet/face/strings.rb
```diff
@@ -96,9 +96,14 @@ def build_generate_options(options = nil, *yard_args)
     if options
       markup = options[:markup]
       generate_options[:markup] = markup if markup
-      json_file = options[:emit_json]
-      generate_options[:json] = json_file if json_file
-      generate_options[:json] = nil if options[:emit_json_stdout]
+      generate_options[:path] = options[:out] if options[:out]
+      generate_options[:stdout] = options[:stdout]
+      format = options[:format] || ""
+      if format.casecmp 'markdown'
```
eputnam: casecmp doesn't do what i thought it did

eputnam:
need to add --emit-json stuff back in and deprecate

eputnam:
need to implement option and raise tags, remove author tag

spec/unit/puppet-strings/markdown_spec.rb
```diff
@@ -214,7 +217,7 @@ class klass (
 
   describe 'rendering markdown to a file' do
     it 'should output the expected markdown content' do
-      Tempfile.open('md') do |file|
+      File.open('/Users/eric.putnam/src/puppet-strings/md.md', 'w') do |file|
```
pegasd: Was this intentional?

eputnam: nope, thank you!

PR #157, opened 2018-02-22
(maint) do not display name if there is nothing to display

eputnam:
If there is a class, type, or function with no information documented but is pulled in by some fluke or tag we don't support, do not give it its own section


scotje:
Will this still show if they have a description comment but no tags?

eputnam:
Yes, that's the @registry[:docstring][:text] from this line:
```ruby
result = @registry[:docstring][:text] ? true : false
```

scotje:
Cool!

PR #159, opened 2018-02-23
Fix return type matching for Puppet functions

pegasd:
This fixes the "Documented return type does not match return type" warning for Puppet functions when the types actually match.


spec/unit/puppet-strings/yard/handlers/puppet/function_handler_spec.rb
```diff
@@ -20,7 +20,7 @@
     let(:source) { 'function foo{' }
 
     it 'should log an error' do
-      expect{ subject }.to output(/\[error\]: Failed to parse \(stdin\): Syntax error at end of file/).to_stdout_from_any_process
+      expect{ subject }.to output(/\[error\]: Failed to parse \(stdin\): Syntax error at end of/).to_stdout_from_any_process
```
pegasd: This slipped through since I was getting a different error message (`at end of input` instead of `at end of file`), hence a failing test.

This should cover both cases.

eputnam:
Thanks @pegasd ! LGTM

PR #162, opened 2018-03-01
(PDOC-224) Handle --emit-json(-stdout) again

ekohl:


lib/puppet/face/strings.rb
```diff
@@ -119,6 +119,8 @@ def build_generate_options(options = nil, *yard_args)
         else
           raise RuntimeError, "Invalid format #{options[:format]}. Please select 'json' or 'markdown'."
         end
+      elsif options[:emit_json] || options[:emit_json_stdout]
+        generate_options[:json] = true
```
eputnam: also need to add `generate_options[:path] = options[:emit_json]` if emit_json is used.

ekohl: updated

eputnam:
@ekohl ++

PR #165, opened 2018-03-20
first edit pass README

jbondpdx:
This edit revises the usage information; stuff about how to document your code is mostly unchanged. I think before we release 2.0, we should maybe get rid of all that stuff and just refer to the style guide? 

@eputnam , there is one question, flagged with TODO, in here, if you could comment on it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,8 @@ Puppet Strings
 ==============
 [![Build Status](https://travis-ci.org/puppetlabs/puppet-strings.png?branch=master)](https://travis-ci.org/puppetlabs/puppet-strings) [![Gem Version](https://badge.fury.io/rb/puppet-strings.svg)](https://badge.fury.io/rb/puppet-strings)
 
-A Puppet command built on [YARD](http://yardoc.org/).
 
-Puppet Strings generates HTML documentation for Puppet extensions written in Puppet and Ruby.
+Puppet Strings creates documentation for Puppet code by reading code and [YARD](http://yardoc.org/)-style code comments and generating HTML, JSON, or Markdown output.
 
 |                |                                                                 |
 | -------------- |---------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,113 +24,123 @@ Puppet Strings generates HTML documentation for Puppet extensions written in Pup
...
-The easiest way to install the `yard` gem is with Puppet itself:
+## Generating documentation with Puppet Strings
 
-For Puppet >= 4 and Puppet Enterprise 2015.2 and later:
+By default, Puppet Strings outputs documentation as HTML, or you can specify JSON or Markdown output instead.
 
-```
-$ puppet resource package yard provider=puppet_gem
-```
+Strings generates reference documentation based on the code and Strings code comments in all Puppet and Ruby source files under the `./manifests/`, `./functions/`, and `./lib/` directories.
```
eputnam: ./types/ , too. https://github.com/puppetlabs/puppet-strings/blob/master/lib/puppet-strings.rb#L5-L8

and ./tasks/ will be added before 2.0.0 as well.

README.md
```diff
@@ -25,113 +24,123 @@ Puppet Strings generates HTML documentation for Puppet extensions written in Pup
...
 ```
-GEM_HOME=/opt/puppet/lib/ruby/gems/1.9.1 puppet resource package redcarpet provider=gem
+$ puppet strings generate 'modules/foo/lib/**/*.rb' 'modules/foo/manifests/**/*.pp' 'modules/foo/functions/**/*.pp' ...
 ```
 
-### 3. Install Puppet Strings
+### Generate documentation in Markdown
 
-Install the [puppet-strings](https://rubygems.org/gems/puppet-strings) gem. To ensure that Strings is installed in the right place, install the gem with Puppet as shown below.
+Strings outputs documentation in Markdown to either an `.md` file or to stdout.
```
eputnam: we may want to leave the part about .md out. It will write the Markdown to any file, .md is not assumed or automatically added...and now i'm seeing that I originally wrote this :) Also, you cannot send it to stdout anymore.

README.md
```diff
@@ -25,113 +24,123 @@ Puppet Strings generates HTML documentation for Puppet extensions written in Pup
...
+   puppet strings generate --format markdown --out docs/INFO.md
+   ```
 
-For Puppet Enterprise 3.8 (Linux):
+### Generate documentation in JSON
 
-```
-GEM_HOME=/opt/puppet/lib/ruby/gems/1.9.1 puppet resource package puppet-strings provider=gem
-```
+Strings can generate a `.json` file or print JSON to stdout. This can be useful for handling or displaying the data with your own custom applications.
```
eputnam: same with the file format here. Maybe it's ok to say .json and .md. My contention is just that it sounds like you can just say `puppet strings --format json --out blah` and expect to get blah.json, when you won't, you'll get blah.

README.md
```diff
@@ -25,113 +24,123 @@ Puppet Strings generates HTML documentation for Puppet extensions written in Pup
...
+## Reference
 
-## Generating documentation with Puppet Strings
 
-The `puppet strings` command processes the `README` file, as well as all Puppet and Ruby source files under the `./manifests/`, `./functions/`, and `./lib/`
-directories, and then creates HTML documentation under the `./doc/` directory.
+The `puppet strings` command generates module documentation based on code and code comments. 
 
-### Generate module documentation
+By default, running `puppet strings` generates HTML documentation for a module into a `docs` directory within that module. To pass any options or arguments, use the `generate` action.
```
eputnam: the default dir is actually `doc/` like above

README.md
```diff
@@ -25,113 +24,123 @@ Puppet Strings generates HTML documentation for Puppet extensions written in Pup
...
+Action   | Description   
+----------------|-------------------------
 
-1. Change directory into the module: `cd /modules/sample-module`.
-2. Run the command: `puppet strings`.
+`generate` | Generates documentation with the specified parameters, including format and output location.
+`server` | Serves documentation for all modules in the [module path](https://docs.puppet.com/puppet/latest/reference/dirs_modulepath.html) locally at `http://localhost:8808`.
 
-### Generate documentation for specific files
+### `puppet strings generate` command reference
```
eputnam: do we want to get into maintaining a command reference in the README? i wonder if we could generate this section :D

README.md
```diff
@@ -25,113 +24,123 @@ Puppet Strings generates HTML documentation for Puppet extensions written in Pup
...
+`--format` | Specifies a format for documentation. | Markdown, JSON    | If not specified, Strings outputs HTML documentation.
+`--out` | Specifies an output location for documentation | A valid directory location and filename.    | If not specified, outputs to default locations depending on format: HTML (`/docs/`), Markdown (main module directory), or JSON (stdout).
+Filenames or directory paths | Outputs documentation for only specified files or directories. | Markdown, JSON.    | If not specified, Strings outputs HTML documentation.
+
+
+## Setting additional options with Rake tasks
+
+You can use Puppet Strings Rake tasks to generate documentation with additional options or to make your generated docs available on [GitHub Pages](https://pages.github.com/).
+
+To specify additional options when generating documentation, use the `puppet strings:generate` rake task. This command behaves exactly as `puppet strings generate`, but allows you to add the following parameters:
```
eputnam: This is interesting that this is from the original documentation. As far as I can tell, you can totally specify most of these options, if not all, in the command line. Though I will say, yard_args on the command line is a little dodgy. I even added a warning to stderr to warn the user of this :)

README.md
```diff
@@ -141,102 +150,112 @@ If you need to specify additional options when generating documentation, use the
...
+* `backtrace`: Enables backtraces for errors when set to `true`.
+* `markup`: Sets the markup language to use. Defaults to `markdown`.
+* `yard_args`: Sets additional arguments to pass to YARD.
 
-To generate JSON documentation to a file, run:
+For example, the task below adds a search pattern, debugs output, backtraces errors, sets the markup language to `markdown`, and passes an additional YARD argument setting the readme file to `README.md`:
 
 ```
-$ puppet strings generate --format json --out documentation.json
+$ rake strings:generate\['**/*{.pp\,.rb}, true, true, markdown, --readme README.md']
```
eputnam: have mercy, rake args tho

README.md
```diff
@@ -141,102 +150,112 @@ If you need to specify additional options when generating documentation, use the
...
 
-Puppet 4 is a typed language, so Puppet Strings automatically documents the parameter types from code. You may optionally include a parameter type in the `@param` tag. Strings will emit a warning and ignore the documented type should it differ from the actual type.
-
-With Puppet 3 code, you must always include the parameter type with the `@param` tag:
-
-```
-# @param first [String] The first parameter for this class.
-# @param second [Integer] The second parameter for this class.
-```
+Puppet Strings automatically documents information such as data types, default values. [TODO: we should be specific; what else?].
```
eputnam: is namevar, return value type (for functions)?

README.md
```diff
@@ -544,7 +512,7 @@ Here are a few other good resources for getting started with documentation:
 
 We love contributions from the community!
 
-If you'd like to contribute to puppet-strings, check out [CONTRIBUTING.md](https://github.com/puppetlabs/puppet-strings/blob/master/CONTRIBUTING.md) to get information on the contribution process.
+If you'd like to contribute to `puppet-strings`, check out [CONTRIBUTING.md](https://github.com/puppetlabs/puppet-strings/blob/master/CONTRIBUTING.md) to get information on the contribution process.
```
eputnam: I wonder if we should once-over CONTRIBUTING also...

README.md
jbondpdx: We don't, no. This is pretty much just where I have it for now. I'll pull it out once I have it in the docs

README.md
jbondpdx: innnteresting... should we take this out? I'd rather just say you can specify the YARD options with the command, if that's true. Unless people are dying to use Rake for stuff?

README.md
jbondpdx: for real

README.md
jbondpdx: Probably?

README.md
eputnam: The help output is actually a little wonky but all those options are definitely available in the CLI

```shell
puppet strings generate --help


USAGE: puppet strings [--format OUTPUT_FORMAT]
[--out PATH]
[--markup FORMAT]
[--emit-json-stdout]
[--emit-json PATH]
[[search_pattern] ...]

Generate documentation from files.

OPTIONS:
  --render-as FORMAT             - The rendering format to use.
  --verbose                      - Whether to log verbosely.
  --debug                        - Whether to log debug information.
  --emit-json PATH               - DEPRECATED: Write JSON representation of the
                                   documentation to the given file.
  --emit-json-stdout             - DEPRECATED: Print JSON representation of the
                                   documentation to stdout.
  --format OUTPUT_FORMAT         - Designate output format, JSON or markdown.
  --markup FORMAT                - The markup format to use for docstring text
                                   (defaults to 'markdown').
  --out PATH                     - Write selected format to PATH. If no path is
                                   designated, strings prints to STDOUT.
```

README.md
```diff
@@ -25,218 +24,209 @@ Puppet Strings generates HTML documentation for Puppet extensions written in Pup
...
-The easiest way to install the `yard` gem is with Puppet itself:
+## Generating documentation with Puppet Strings
 
-For Puppet >= 4 and Puppet Enterprise 2015.2 and later:
+By default, Puppet Strings outputs documentation as HTML, or you can specify JSON or Markdown output instead.
 
-```
-$ puppet resource package yard provider=puppet_gem
-```
+Strings generates reference documentation based on the code and Strings code comments in all Puppet and Ruby source files under the `./manifests/`, `./functions/`, `./lib/`, `./types/``./tasks/` directories.
```
eputnam: there's an extra backtick after ./types/ and no comma

README.md
```diff
@@ -2,9 +2,8 @@ Puppet Strings
 ==============
 [![Build Status](https://travis-ci.org/puppetlabs/puppet-strings.png?branch=master)](https://travis-ci.org/puppetlabs/puppet-strings) [![Gem Version](https://badge.fury.io/rb/puppet-strings.svg)](https://badge.fury.io/rb/puppet-strings)
 
-A Puppet command built on [YARD](http://yardoc.org/).
 
-Puppet Strings generates HTML documentation for Puppet extensions written in Puppet and Ruby.
+Puppet Strings creates documentation for Puppet code by reading code and [YARD](http://yardoc.org/)-style code comments and generating HTML, JSON, or Markdown output.
```
eputnam: Hmmm, it creates docs for Ruby code, also. 
What about something like:

> Puppet Strings is a Puppet face that extends [YARD](http://yardoc.org) to read Puppet code and YARD-style comments, then generate HTML, JSON, or Markdown documentation.

Mentioning it as a YARD extension implies Ruby functionality as well.

README.md
```diff
@@ -25,218 +24,209 @@ Puppet Strings generates HTML documentation for Puppet extensions written in Pup
...
 
-Puppet Strings supports the Markdown format and automatically sets the YARD `markup` option to `markdown`.
+## Viewing generated documentation
 
-## Generating documentation with Puppet Strings
+Strings generates documentation as HTML, JSON, or Markdown within the module for which you are generating documentation.
 
-The `puppet strings` command processes the `README` file, as well as all Puppet and Ruby source files under the `./manifests/`, `./functions/`, and `./lib/`
-directories, and then creates HTML documentation under the `./doc/` directory.
+By default, Strings outputs documentation as HTML in a `/docs/` folder in the module. If you generate Markdown documentation with Strings, it outputs a `REFERENCE.md` file in the main directory of the module.
```
eputnam: ./doc/ folder :)

README.md
```diff
@@ -394,8 +384,8 @@ To document a function in the Puppet 3 API, use the `doc` option to `newfunction
 ```ruby
 Puppet::Parser::Functions.newfunction(:example, doc: <<-DOC
 Documentation for an example 3.x function.
```
eputnam: I just noticed this is a Puppet 3 example. Kill with fire. Are we just going to get rid of all these examples eventually and point to the style guide anyway?

jbondpdx: I left this in because I had the idea that someone might need to document a legacy function. Still kill it?

jbondpdx: Oh, and yes, we are going to get rid of them and point to the style guide Real Soon Now.

eputnam: ah ok, legacy is fair.

README.md
jbondpdx: dangit!

README.md
jbondpdx: I tried something else, because I didn't want to talk about faces, and I want to talk about what it *does* rather than what it *is*. See what you think.

PR #170, opened 2018-03-30
(PDOC-233) markdown whitespace fixes

JohnLyman:
- Adds blank line padding around header elements so they will render correctly with other markdown variants
- Minor updates to make whitespace consistent

I wasn't sure how to add tests for this to prevent future updates from breaking rendering.  Someone would need to know to add blank lines around headers in the output fixtures if/when they are updated.  I thought of adding a test that would read the fixture file and check the linespace padding, but if a new fixture is added (like one was recently), someone would also need to know to add a test against the new fixture.  Testing the test fixture itself is not typical and didn't feel right.

### DIFF
--- a/lib/puppet-strings/markdown/templates/classes_and_defines.erb
+++ b/lib/puppet-strings/markdown/templates/classes_and_defines.erb
@@ -24,8 +24,10 @@
 <% end -%>
 <% if examples -%>
 #### Examples
+
 <% examples.each do |eg| -%>
 ##### <%= eg[:name] %>
+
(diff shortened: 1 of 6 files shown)


eputnam:
this is awesome @JohnLyman ! Thanks!

It's actually fairly common practice at Puppet to have spacing around headers, I was just being lazy in this case because I didn't think it was a big deal. I was wrong! Anyway, I'll at least try to look out for it until we come up with a way to maybe keep an eye on it.

PR #174, opened 2018-04-20
(PDOC-240) add handling for :array node type in rsapi_handler

eputnam:
The 'features' key in the resource api expects an array of feature names. Currently, if you run strings on a resource api type that specifies features, it falls over because it doesn't know what to do with an AST node of type :array. Further, the method that throws the error is looking for an undefined variable called 'parameters'. This removes the odd conditional that calls 'parameters' and adds a method that returns an array of the array contents that may be passed to features. Features are not currently sent to the renderer, but now Strings will be able to handle any array it comes across in an rsapi type.


DavidS:
:+1: 

thanks!

PR #181, opened 2018-06-04
(PDOC-255) markdown table of contents update

eputnam:
Markdown output does not currently distinguish between the group headings for the table of contents and group headings in the actual content. This change adds a 'Table of Contents' heading and changes group headings to bold text to make it much more clear that you're looking at the table of contents.


eputnam:
Sample output with these changes:
screen shot 2018-06-03 at 8 22 31 pm

kris-bosland:
I don't see a button for merge - I assume it's a permission problem but I don't know where to look for the permissions.

PR #183, opened 2018-06-13
(PDOC-252) Add describe features to puppet-strings face

kris-bosland:
Hey @eputnam , here is that PR that I was talking about.  LMK when you have had some time to look at this, and then I would like to talk about how I can pull in the data I need to finish describe features.  Thanks!

### DIFF
--- a/lib/puppet-strings.rb
+++ b/lib/puppet-strings.rb
@@ -53,13 +53,17 @@ module PuppetStrings
     YARD::CLI::Yardoc.run(*args)
 
     # If outputting JSON, render the output
-    if options[:json]
+    if options[:json] && !options[:describe]
       render_json(file)
     end
 
     # If outputting Markdown, render the output
     if options[:markdown]
       render_markdown(file)
+    end
+
+    if options[:describe]
+      render_describe(options[:describe_types], options[:describe_list], options[:providers])
     end
   end
 
@@ -75,6 +79,11 @@ module PuppetStrings
   def self.render_markdown(path)
     require 'puppet-strings/markdown'
     PuppetStrings::Markdown.render(path)
+  end
+
+  def self.render_describe(describe_types, list = false, providers = false)
+    require 'puppet-strings/describe'
+    PuppetStrings::Describe.render(describe_types, list, providers)
   end
 
   # Runs the YARD documentation server.
--- a/lib/puppet-strings/describe.rb
+++ b/lib/puppet-strings/describe.rb
@@ -1,0 +1,66 @@
+require 'json'
+require 'puppet-strings/json'
+
+# The module for command line documentation related functionality.
+module PuppetStrings::Describe
+  # Renders requested types or a summarized list in the current YARD registry to STDOUT.
+  # @param [Array] describe_types The list of names of the types to be displayed.
+  # @param [bool] list Create the summarized list instead of describing each type.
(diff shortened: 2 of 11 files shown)


lib/puppet-strings.rb
```diff
@@ -53,14 +53,18 @@ def self.generate(search_patterns = DEFAULT_SEARCH_PATTERNS, options = {})
...
+    if options[:json] && !options[:describe]
       render_json(file)
     end
 
     # If outputting Markdown, render the output
     if options[:markdown]
       render_markdown(file)
     end
+
+    if options[:describe]
```
eputnam: If we want to have `--describe` as a generate option, we'll need to register it in the face https://github.com/puppetlabs/puppet-strings/blob/master/lib/puppet/face/strings.rb#L7

on further examination this is just something that the action passes to generate, i guess.

lib/puppet/face/strings.rb
```diff
@@ -81,6 +81,65 @@
...
+      #TODO: Set up search_patterns and whatever else needed to collect data for describe - currently missing some
+      #          manifests/**/*.pp
+      #          functions/**/*.pp
+      #          tasks/*.json
+      #          plans/*.pp
+      search_patterns = %w(
+        types/**/*.pp
+        lib/**/*.rb
+      )
+      PuppetStrings::generate(
```
eputnam: I wonder about trying to create a `PuppetStrings::describe` method separate from `PuppetStrings::generate`. We may be able to cut some of the cruft out to reduce load time.

lib/puppet-strings/describe.rb
```diff
@@ -0,0 +1,60 @@
...
+require 'puppet-strings/json'
+
+# The module for command line documentation related functionality.
+module PuppetStrings::Describe
+  # Renders requested types or a summarized list in the current YARD registry to STDOUT.
+  # @param [Array] describe_types The list of names of the types to be displayed.
+  # @param [bool] list Create the summarized list instead of describing each type.
+  # @param [bool] providers Show details of the providers.
+  # @return [void]
+  def self.describe(describe_types = [], list = false, providers = false)
```
eputnam: sort of a nit, but we've been calling the methods that create the final output `render`

lib/puppet-strings.rb
kris-bosland: Yes, I am just using it to carve out a space in the generate options, since I expect most of the describe options to not be applicable to the other modes.  Probably a little awkward style wise.

lib/puppet/face/strings.rb
kris-bosland: Any suggestions for stuff we don't need for describe @eputnam ?

eputnam: I forget now what I was thinking here. I guess that describe doesn't need to look at everything we list in generate normally? But as for what that is, I can't remember.

kris-bosland: Yeah, I guess the search_patterns I set up here are much less than the regular generate.  I think for now I will focus on filling in missing data, then I will come back and look for performance improvements.

kris-bosland:
Hey @eputnam , I think this is ready for merge - it isn't a complete reimplementation of describe, but I think it makes more sense to do this work in several pull requests.  This does pull in the missing pieces of the file type that we were looking at back in June.

PR #186, opened 2018-06-26
Avoid clashes with PSH's build task on release

DavidS:
The (module-)`build` task from puppetlabs_spec_helper is conflicting
with the (gem-)`build` task from bundler.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,5 @@
 require 'bundler/gem_tasks'
-require 'puppetlabs_spec_helper/rake_tasks'
+#require 'puppetlabs_spec_helper/rake_tasks'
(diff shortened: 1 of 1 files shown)


DavidS:
This is a bit of a doozy, since I've released the 2.1.0 off this commit, but I'm not sure that there is anything in puppetlabs_spec_helper/rake_tasks that we need to keep.

eputnam:
Looks like we rely on PSH pretty heavily for testing Strings, unfortunately. We'll have to figure out another way to run tests or do some more modularization of rake tasks in PSH.

DavidS:
@eputnam thanks for catching that. Here's the fix for the `spec` task, this should now actually run the tests, and we can merge the release back into master.

PR #187, opened 2018-07-10
(MAINT) Bump rubocop dependency to address CVE-2017-8418

scotje:


scotje:
@eputnam I don't think so, rubocop seems to just be a dev dependency?

scotje:
Wouldn't hurt, although since this is itself a gem, runtime dependencies should be in the .gemspec so basically everything in the Gemfile could be considered a development dependency.

PR #188, opened 2018-09-14
(PDOC-265) Add examples to function reference docs

ekohl:


alexjfisher:
Can you add a test for 4.x API functions too?
Somewhere in here? https://github.com/puppetlabs/puppet-strings/blob/00a0f0e16270b0ce89f8d705847a3ec4dab237d2/spec/unit/puppet-strings/markdown_spec.rb#L118-L125

ekohl:
@alexjfisher like this?

alexjfisher:
@ekohl Not sure...

Where are the examples supposed to go? Above the function, or above the dispatch (or both!)?  Style-guide suggests above the dispatch. https://puppet.com/docs/puppet/5.5/puppet_strings_style.html#example-ruby-function-with-one-potential-return-type

alexjfisher:
ping @hlindberg

ekohl:
I did consider also adding examples to the dispatch functions. That might also make sense for when you use multiple signatures

hlindberg:
It would be nice to be able to write examples above the dispatchers. Sometimes you don't want to do that as the main part shows a progression of examples using different signatures. So, both is preferable IMO.

ekohl:
Updated. It now parses examples within dispatchers/signatures as well. This PR also contains 2 cleanup commits that made it easier for me to understand what was going on but not strictly needed.

lib/puppet-strings/yard/handlers/ruby/function_handler.rb
```diff
@@ -17,6 +17,7 @@ class PuppetStrings::Yard::Handlers::Ruby::FunctionHandler < PuppetStrings::Yard
     required_block_param
     optional_block_param
     return_type
+    example
```
eputnam: `example` isn't a parameter method, is it? you can remove this.

ekohl: You're right, it was a debug leftover.

ekohl:
Anything holding back a merge?

PR #190, opened 2018-11-18
Correction in comments

alexharv074:
The #defaults method actually returns a Hash, not an Array.


scotje:
Thanks for the correction!

PR #191, opened 2018-11-21
Add site to default search path

alexharv074:
The recommended control-repo has a site directory in the top-level,
i.e. https://github.com/puppetlabs/control-repo/tree/production/site

This patch adds site to the default search path.

### DIFF
--- a/lib/puppet-strings.rb
+++ b/lib/puppet-strings.rb
@@ -2,6 +2,7 @@
 module PuppetStrings
   # The glob patterns used to search for files to document.
   DEFAULT_SEARCH_PATTERNS = %w(
+    site/**/*.pp
     manifests/**/*.pp
     functions/**/*.pp
     types/**/*.pp


binford2k:
Does this mean using `puppet-strings` to document an entire control repository? I thought it was intended to document a single module.

scotje:
Yeah, I'm not sure documenting a whole control-repo is a use-case for `puppet-strings` that we had considered yet. I'm not conceptually opposed to it though, @alexharv074 have you already started using this locally? What has your experience been like so far?

alexharv074:
@scotje , @binford2k Yes. Puppet Strings does appear to be intended only to document a single module. I haven't finished my work to document a Control Repo using Puppet Strings, but you can see my work in progress [here](https://github.com/alexharv074/elk/commit/eef7206d857af5e7bc0b02bf7c4c44f27f43a119). After applying the patch in this pull request, I could generate Markdown from classes inside a site directory.

Note that I used Puppet Strings [here](https://github.com/alexharv074/puppet-firewall_multi/blob/master/.README.erb#L3-L34) and [here](https://github.com/alexharv074/puppet-firewall_multi/blob/master/.README.erb#L104).

See also my blog post [here](https://alexharv074.github.io/2018/11/09/generating-markdown-from-source-code-in-erb.html) on how I generally use ERB to generate nearly all my README files. Since I think all duplication in docs is bad, I do think Puppet Strings's job should expand from just documenting modules.

binford2k:
Hey @alexharv074, we've talked about this a bit and decided that we want to keep the scope to documenting a single module only. I like the way you're invoking the methods with a search path, that seems like a pretty elegant way to solve your problem. It seems like that could be part of a `Rakefile` in your control repo itself. Thanks for the idea!

alexharv074:
@binford2k Is there a way I can refactor to somehow override that search pattern?

PR #192, opened 2019-01-11
(maint) Implement a strings:generate:reference task

ekohl:
This makes it easy to generate just a reference


ekohl:
The build is failing due to a Bundler issue, probably because of Bundler 2.0.

PR #197, opened 2019-05-31
Update README.md

cfawcett82:
Update install strings to use Puppets internal gem instead of system wide gem
This was required for use on Mac OSX 10.14.x

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Puppet Strings generates documentation for Puppet code and extensions written in
 
 ### Install Puppet Strings
 
-1. Install the YARD gem by running sudo /opt/puppetlabs/puppet/bin/gem install `yard`
+1. Install the YARD gem by running `sudo /opt/puppetlabs/puppet/bin/gem install yard`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,8 +30,8 @@ Puppet Strings generates documentation for Puppet code and extensions written in
 
 ### Install Puppet Strings
 
-1. Install the YARD gem by running `gem install yard`
-1. Install the `puppet-strings` gem by running `gem install puppet-strings`
+1. Install the YARD gem by running sudo /opt/puppetlabs/puppet/bin/gem install `yard`
```
gguillotte: Probably want the Markdown backtick before sudo:

```
 1. Install the YARD gem by running `sudo /opt/puppetlabs/puppet/bin/gem install yard`
```

> 1. Install the YARD gem by running `sudo /opt/puppetlabs/puppet/bin/gem install yard`

cfawcett82: good catch @gguillotte, updated.

scotje:
Thanks for the contribution @cfawcett82!

I've modified this a bit and opened a new PR (#198) to reflect the fact that there are two different paths for installation depending on how you have installed Puppet itself.

PR #204, opened 2019-06-25
(maint) Use Litmus for acceptance testing.

glennsarti:
Previously puppet-strings depended on Beaker which could only run on hosted CI
platform, however with the new Litmus product we are able to move the acceptance
testing to a per PR check via automated CI (Travis).  This commit:

* Updates the Gemfile to use Litmus instead of Beaker.
* Updates the RakeFile and fixtures to pull in the required tasks and needed
  Puppet modules.  Note that the empty Symlink array is required to stop
  symlinks being created for the source directory as it's not needed.
* Migrated acceptance tests to Litmus format. Note that two tests are failing
  and are currently skipped.
* Updated the module metadata to be complete and actually be able to built into
  a tarball.
* Updates Travis CI configuration to run the acceptance tests.


.travis.yml
```diff
@@ -24,5 +24,18 @@ matrix:
...
+    rvm: 2.5
+    before_script:
+    - bundle exec rake 'litmus:provision[docker, centos:7]'
+    - bundle exec rake 'litmus:install_agent[puppet6]'
+    - bundle exec rake litmus:install_module_fixtures
+    - bundle exec rake litmus:install_gems
+    script:
+    - bundle exec rake litmus:acceptance:parallel
+    services: docker
+    sudo: required
```
tphoney: we should this as an opportunity to remove the unsupported puppet 4 tests.

Rakefile
```diff
@@ -1,5 +1,6 @@
+require 'puppet_litmus/rake_tasks' if Bundler.rubygems.find_name('puppet_litmus').any?
```
tphoney: this entire file, gives me feels. But it is needed :(

spec/spec_helper_acceptance.rb
```diff
@@ -1,28 +1,58 @@
...
+      endpoint: endpoint,
+      operation_timeout: 300,
+    }
+
+    winrm = WinRM::Connection.new opts
+    Specinfra.configuration.winrm = winrm
   end
 end
+
+require 'spec_helper_acceptance_local' if File.file?(File.join(File.dirname(__FILE__), 'spec_helper_acceptance_local.rb'))
```
tphoney: any reason this is at the bottom of the file ?

spec/acceptance/running_strings_generate_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+describe 'Generating module documentation using generate action' do
+  before :all do
+    # TODO: Linux only
+    @sut_work_dir = PuppetLitmus::Serverspec.run_shell("pwd").stdout.chomp
+
+    test_module_path = sut_module_path(/Module test/)
+    PuppetLitmus::Serverspec.run_shell("puppet strings generate \"#{test_module_path}/**/*.{rb,pp}\"")
+  end
+
+  def expect_file_contain(path, expected_contents)
```
tphoney: it would be good to move this helper function out to spec_helper_acceptance_local.rb

spec/acceptance/emit_json_options_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  [
+    { :title => '--format json and --out', :cmd_line => '--format json --out "TMPFILE"' },
+    { :title => '--emit-json', :cmd_line => '--emit-json "TMPFILE"' },
+  ].each do |testcase|
+    it "should write JSON to a file when using #{testcase[:title]}" do
+      tmpfile = File.join(@remote_tmp_path, 'json_output.json')
+      cmd = "puppet strings generate #{testcase[:cmd_line].gsub('TMPFILE', tmpfile)} \"#{@test_module_path}/lib/puppet/parser/functions/function3x.rb\""
+      PuppetLitmus::Serverspec.run_shell(cmd)
+      output = JSON.parse(file(tmpfile).content)
```
tphoney: if you use serverspec, you will get many helpers for things like this.  
`describe file("c:/#{test_dir_path}/#{fake_name}") do    
      its(:content) { should match /#{test_file_contents}/ }   
end`

Rakefile
```diff
@@ -23,38 +24,106 @@ task :validate do
...
+    tmp_path = '/tmp/'
+    run_local_command("bundle exec bolt file upload #{gem_tar} #{tmp_path}#{File.basename(gem_tar)} --nodes #{target_string} --inventoryfile inventory.yaml")
+
+
+    # Install dependent gems
+    install_remote_gem('yard', target_nodes, inventory_hash)
+    install_remote_gem('rgen', target_nodes, inventory_hash)
+    # Install puppet-strings
+    install_remote_gem(tmp_path + File.basename(gem_tar), target_nodes, inventory_hash)
+    puts 'Installed'
```
scotje: Is all this code needed because Litmus doesn't have a built-in way to build/install gems on the SUT instead of a module?

Rakefile
```diff
@@ -1,5 +1,6 @@
+require 'puppet_litmus/rake_tasks' if Bundler.rubygems.find_name('puppet_litmus').any?
+require 'puppetlabs_spec_helper/rake_tasks'
```
scotje: What tasks are we using from PSH?

spec/acceptance/generate_markdown_spec.rb
```diff
@@ -1,9 +1,11 @@
 require 'spec_helper_acceptance'
-require 'util'
-
-include PuppetStrings::Acceptance::Util
 
 describe 'Generating Markdown' do
+  before(:all) do
+    @test_module_path = sut_module_path(/Module test/)
+    @remote_tmp_path = sut_tmp_path
+  end
```
scotje: Unless these are particularly expensive or need to be shared between examples, I would prefer these to be `let` or `before(:each)`

spec/spec_helper_acceptance.rb
scotje: @glennsarti can you elaborate on why we need the `_local` version of `spec_helper_acceptance`? Does Litmus do something special with that? I don't think we are managing these files from a template with (e.g. `pdk update`) so it would seem fine to just have a few helpers directly in the main `spec_helper_acceptance`?

spec/acceptance/generate_markdown_spec.rb
```diff
@@ -31,19 +33,15 @@ class { "test": }
...
   EOF
 
-  it 'should render Markdown to stdout when using --format markdown and --stdout' do
-    test_module_path = get_test_module_path(master, /Module test/)
-    on master, puppet('strings', 'generate', '--format markdown', "#{test_module_path}/manifests/init.pp") do
-      output = stdout.chomp
-      expect(JSON.parse(output)).to eq(expected)
-    end
+  it 'should render Markdown to stdout when using --format markdown' do
+    skip('This test is broken. Does not output to STDOUT by default.')
```
scotje: Yep, looking at the code I see why. I'll file a ticket if there isn't one already.

.travis.yml
```diff
@@ -24,5 +24,18 @@ matrix:
...
+    bundler_args: ""
+    dist: trusty
+    rvm: 2.5
+    before_script:
+    - bundle exec rake 'litmus:provision[docker, centos:7]'
+    - bundle exec rake 'litmus:install_agent[puppet6]'
+    - bundle exec rake litmus:install_module_fixtures
+    - bundle exec rake litmus:install_gems
+    script:
+    - bundle exec rake litmus:acceptance:parallel
```
scotje: Is there an easy way to run the acceptance tests locally or do we just have to run all 5 rake commands in order? `rake litmus:acceptance:localhost` doesn't seem to find any examples.

glennsarti: >Is there an easy way to run the acceptance tests locally

Currently no. 

> do we just have to run all 5 rake commands in order? 

What I'm missing is a README update to explain how to run the Acceptance Tests and why.

.travis.yml
glennsarti: Maybe, but that seems out of scope for this particular PR.

Rakefile
glennsarti: Litmus runs `spec_prep` which comes from PSH

Rakefile
glennsarti: Agreed, but I didn't see another clean way to do it.  Particularly as Puppet-Strings exposes public Rake Tasks too.  Didn't want to pollute those.

That said, I tried to make my additions "modular" so when Litmus does support this stuff it'll be easier to port (I hope!)

Rakefile
glennsarti: Correct. Litmus, right now, is purely a module content testing tool. Not a generic testing tool.  As Litmus progresses these kinds of helpers would become a Litmus plugin/extension/module or whatever we call it. As stated above, I tried keep these tasks in the same spirit as the original Litmus code. Ref - https://github.com/puppetlabs/puppet_litmus/blob/master/lib/puppet_litmus/rake_tasks.rb#L252-L296

spec/acceptance/emit_json_options_spec.rb
glennsarti: Unfortunately I can't because the file path is determined inside a test case (Second pass of rspec).
I can't hardcode the file paths because these tests should be platform and provisioning agnostic e.g. A vagrant RHEL VM could have a different `pwd` than a docker CentOS container or a HyperV Windows VM.

spec/acceptance/generate_markdown_spec.rb
glennsarti: I did originally use `:each` and it works fine

* They're not too expensive on Linux Docker containers, but once this tests on Windows over WinRM, they will take a while (5+ seconds per call IIRC)

* The values don't change during the lifetime of the SUT so I didn't see the point of running the same commands multiple times.

I don't have a hard opinion either way. Both approaches work, and this is faster.  That was all.

spec/acceptance/running_strings_generate_spec.rb
glennsarti: Yeah.  If it's needed outside of this file sure. Currently it's not.

Personally, I prefer to keep "magic" as close to the tests as possible to stop people chasing over different files e.g. this project used `/spec/acceptance/lib/*` and `/spec/spec_acceptance_helper.rb` to hold it's helper code.

spec/spec_helper_acceptance.rb
glennsarti: > any reason this is at the bottom of the file ?
So that local `acceptance_local` can do configuration AFTER the target nodes have a transport ready for use, or override Litmus behaviour.  The default location (e.g. MOTD - https://github.com/puppetlabs/puppetlabs-motd/blob/055cd0668009f47a481f15cd10c041bd7641c8f9/spec/spec_helper_acceptance.rb#L3-L6) means you can't config nodes

Currently this functionality is not required, but could be in the future. Happy to move it up the top.

glennsarti: >  can you elaborate on why we need the _local version of spec_helper_acceptance

Strictly speaking we don't. I was trying to keep this file as close as possible to what Litmus is expecting due to it being developed so quickly and making it easier to update in the future.  There was no consideration of `pdksync`

.travis.yml
glennsarti: Added a README change.

Rakefile
glennsarti: TODO: Can I just import spec_prep?

Rakefile
```diff
@@ -23,38 +24,106 @@ task :validate do
...
-    puts "INFO: adding 'type=#{install_type}' to host config"
-    target += "{type=#{install_type}}"
+    target_string = if args[:target_node_name].nil?
+                      'all'
+                    else
+                      args[:target_node_name]
+                    end
+    # TODO: Currently this is Linux only
+    tmp_path = '/tmp/'
+    run_local_command("bundle exec bolt file upload #{module_tar} #{tmp_path}#{File.basename(module_tar)} --nodes #{target_string} --inventoryfile inventory.yaml")
```
glennsarti: SHould this be using the BOlt:Runner namespace stuff?

Rakefile
scotje: It looks like maybe `require 'puppetlabs_spec_helper/tasks/fixtures'` would maybe pick up just `spec_prep`, `spec_clean`, and `spec_clean_symlinks`. Not sure if it depends on stuff rom the parent file though.

glennsarti: Looks good. Will try.

Rakefile
glennsarti: It probably should but .. the current litmus rake helpers use this style e.g.

https://github.com/puppetlabs/puppet_litmus/blob/45feb78c2df91b2110900e63baf89693e9dbd511/lib/puppet_litmus/rake_tasks.rb#L283

Rakefile
glennsarti: Took your advice and fixed.

PR #205, opened 2019-06-29
(PDOC-283) Fix namespaced symbols

glennsarti:
Previously in Yard 0.9.19 dyna_symbols (symbols represented as strings e.g.
:'something') were converted to their unqualified symbol name e.g. :something.
However YARD treated this as a bug and in Yard 0.19.20 dynasymbols are now their
original source.  This commit modifies the base ruby handler to always convert
dyna_symbols to their plain name, as Puppet uses this for actual documentation
and Puppet Code.  For example the fully qualified function 'foo::bar' will
appear in the ruby function definition as `.. :'foo::bar'` however in the Puppet
documentation the name should appears as `foo::bar`.

This commit also adds tests for this scenario as there were no existing tests
for this.

### DIFF
--- a/lib/puppet-strings/yard/handlers/ruby/base.rb
+++ b/lib/puppet-strings/yard/handlers/ruby/base.rb
@@ -18,7 +18,13 @@ class PuppetStrings::Yard::Handlers::Ruby::Base < YARD::Handlers::Ruby::Base
     when :label
       node.source[0..-2]
     when :dyna_symbol
-      node.source
+      # YARD 0.9.20 changed how dyna_symbols are represented
+      # https://github.com/lsegal/yard/commit/225ded9ef38c6d2be5a3b0fc7effbc7d6644768d
+      if yard_version >= Gem::Version.new('0.9.20')
+        node.source[2..-2]
+      else
+        node.source
+      end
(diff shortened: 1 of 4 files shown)


lib/puppet-strings/yard/handlers/ruby/base.rb
```diff
@@ -18,7 +18,13 @@ def node_as_string(node)
     when :label
       node.source[0..-2]
     when :dyna_symbol
-      node.source
+      # YARD 0.9.20 changed how dyna_symbols are represented
+      # https://github.com/lsegal/yard/commit/225ded9ef38c6d2be5a3b0fc7effbc7d6644768d
+      if yard_version >= Gem::Version.new('0.9.20')
+        node.source[2..-2]
```
binford2k: This seems like a really fragile way to do this, but I trust your judgement that there's not a better way.

glennsarti: Yeah I wasn't super happy with it.  But it's no worse than the examples above or below 🤷‍♂️ The fragility is also why I add tests so at least we should be able to detect when it goes bad.

PR #215, opened 2019-10-20
(PDOC-295) Add @enum tag support for Enum data types

seanmil:
`Enum` data types would benefit from the ability to add extended
documentation on each option similarly to the `@option` tag
for hashes. The `@option` type is not a good fit for `Enum` data
type parameters as it requires data types to be specified,
which doesn't make sense in the context of an `Enum` type.

This patch adds a new `@enum` tag which behaves like @option,
but does not accept the data type parameter that `@option` does.

### DIFF
--- a/lib/puppet-strings/markdown/base.rb
+++ b/lib/puppet-strings/markdown/base.rb
@@ -111,12 +111,25 @@ module PuppetStrings::Markdown
       select_tags('option')
     end
 
+    # @return [Array] enum tag hashes
+    def enums
+      select_tags('enum')
+    end
+
     # @param parameter_name
     #   parameter name to match to option tags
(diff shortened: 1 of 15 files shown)


scotje:
I think the acceptance test failure is due to a PDK "missing require" bug that should be fixed with the new PDK release this week. After that I'm 👍 to merge this.

scotje:
Re-kicking CI now that PDK 1.14.1 gem has been released.

scotje:
Fresh new failures now, hoping https://github.com/puppetlabs/puppet-strings/pull/217 will resolve

scotje:
Rebased on top of #217.

scotje:
Hooray, CI passing now!

PR #218, opened 2019-11-29
Add missing HTML output support for enum tag

seanmil:
In #215 support for the new `enum` tag was added. Unfortunately, all of the testing was with markdown generation and it escaped me that HTML output was not automatically included. This patch adds the HTML output generation for the `enum` tag which should have been included in #215


spec/fixtures/acceptance/modules/test/manifests/init.pp
```diff
@@ -7,9 +7,13 @@
 #
 # @param package_name The name of the package
 # @param service_name The name of the service
+# @param enum
```
scotje: Should this be:

```suggestion
# @param myenum
```

PR #232, opened 2020-05-01
Add checks to resource_type handler and code objects

### LINKED ISSUE #229: exec type in generated docs missing attributes `creates`, `onlyif`
## Describe the Bug
In generated Puppet reference documentation, the `exec` type is missing information for the `creates` and `onlyif` attributes

## Expected Behavior
Generated docs should include both attribute descriptions, and the attributes should also appear at the code block at the top of the page, without manual intervention. (See https://puppet.com/docs/puppet/latest/types/exec.html , the attributes were added manually after generation)

## Steps to Reproduce
1. Within the puppet-docs repo, run the internal docs generation command: `bundle exec rake references:puppet VERSION=6.14.0`
2. Diff the generated doc from puppet-docs/reference_output/puppet/types/exec.md against the existing documentation in puppet-docs/source/puppet/latest/types/exec.md

## Environment
cloned puppet-docs repo, *nix

## Additional Context
The generation command retrieves a vendored copy of the Puppet codebase; there's a setting you can change to have it use a Puppet codebase on your machine, if you need to test documentation changes sourced from the Puppet code itself. (I don't think this issue will require that, though.)

scotje:
Resolves #229

### DIFF
--- a/lib/puppet-strings/markdown/resource_type.rb
+++ b/lib/puppet-strings/markdown/resource_type.rb
@@ -12,11 +12,28 @@ module PuppetStrings::Markdown
     end
 
     def properties
-      @registry[:properties]
+      return nil unless @registry[:properties]
+
+      @registry[:properties].sort_by { |p| p[:name] }
+    end
+
+    def checks
+      return nil unless @registry[:checks]
+
+      @registry[:checks].sort_by { |p| p[:name] }
+    end
+
+    # "checks" (such as "onlyif" or "creates") are another type of property
+    def properties_and_checks
+      return nil if properties.nil? && checks.nil?
+
+      ((properties || []) + (checks || [])).sort_by { |p| p[:name] }
     end
 
     def parameters
-      @registry[:parameters]
+      return nil unless @registry[:parameters]
+
+      @registry[:parameters].sort_by { |p| p[:name] }
     end
 
     def regex_in_data_type?(data_type)
--- a/lib/puppet-strings/markdown/templates/resource_type.erb
+++ b/lib/puppet-strings/markdown/templates/resource_type.erb
@@ -44,12 +44,12 @@
 
 <% end -%>
 <% end -%>
-<% if properties -%>
+<% if properties_and_checks -%>
 #### Properties
 
 The following properties are available in the `<%= name %>` <%= @type %>.
 
-<% properties.each do |prop| -%>
+<% properties_and_checks.each do |prop| -%>
 ##### `<%= prop[:name] %>`
 
 <% if prop[:values] -%>
--- a/lib/puppet-strings/yard/code_objects/type.rb
+++ b/lib/puppet-strings/yard/code_objects/type.rb
@@ -73,6 +73,9 @@ class PuppetStrings::Yard::CodeObjects::Type < PuppetStrings::Yard::CodeObjects:
   class Property < Parameter
   end
 
+  class Check < Parameter
+  end
+
   # Represents a resource type feature.
   class Feature
     attr_reader :name, :docstring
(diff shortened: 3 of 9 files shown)


scotje:
cc @jbondpdx @eputnam

lib/puppet-strings/markdown/resource_type.rb
```diff
@@ -12,11 +12,28 @@ def render
...
+      @registry[:properties].sort_by { |p| p[:name] }
+    end
+
+    def checks
+      return nil unless @registry[:checks]
+
+      @registry[:checks].sort_by { |p| p[:name] }
+    end
+
+    # "checks" (such as "onlyif" or "creates") are another type of property
```
binford2k, line 26: is new check _really_ a totally new type of property invented only for the exec type?!

https://github.com/puppetlabs/puppet/search?q=newcheck&unscoped_q=newcheck

scotje: 🤷‍♂️ 

Is it also available for custom resource types and providers?

binford2k: Nope. It's defined and used in only that one type.

@joshcooper is `newcheck` something that we might want to elevate to be part of the actual T&P API? Or could we refactor these psuedoparameters into actual params? My concern is that with this being special-cased is that other parts of the ecosystem will almost certainly not account for these either.

(We'll obviously still have to handle this in strings for backwards compatibility)

binford2k: Or maybe it doesn't matter. It's clearly _also_ defining parameters and just doesn't catch it because it's static analysis. The type object will have a param defined.

https://github.com/puppetlabs/puppet/blob/b6d0627415fcf8d25c81e044acf95faf00880a22/lib/puppet/type/exec.rb#L70-L75

joshcooper: `newcheck` is just for the `exec` type, however, the net result is `:creates`, etc are metaprogrammed parameters. I'm not familiar with how strings introspects the type/property/parameter, but it's possible to access the documentation for all exec parameters (which include properties):

```
irb(main):001:0> require 'puppet'
=> true
irb(main):002:0> Puppet::Type.type(:exec)
=> Puppet::Type::Exec
irb(main):003:0> Puppet::Type.type(:exec).parameters
=> [:command, :path, :user, :group, :cwd, :logoutput, :refresh, :environment, :umask, :timeout, :tries, :try_sleep, :refreshonly, :creates, :unless, :onlyif, :provider]
irb(main):004:0> Puppet::Type.type(:exec).paramclass(:creates)
=> Puppet::Type::Exec::ParameterCreates
irb(main):005:0> Puppet::Type.type(:exec).paramclass(:creates).doc
=> "A file to look for before running the command. The command will\nonly run if the file **doesn't exist.**\n\nThis parameter doesn't cause Puppet to create a file; it is only\nuseful if **the command itself** creates a file.\n\n    exec { 'tar -xf /Volumes/nfs02/important.tar':\n      cwd     => '/var/tmp',\n      creates => '/var/tmp/myfile',\n      path    => ['/usr/bin', '/usr/sbin',],\n    }\n\nIn this example, `myfile` is assumed to be a file inside\n`important.tar`. If it is ever deleted, the exec will bring it\nback by re-extracting the tarball. If `important.tar` does **not**\nactually contain `myfile`, the exec will keep running every time\nPuppet runs.\n\n"
```

scotje:
Based on how strings analyzes things, I think the current approach is still the best at the moment, so I'm going to merge this. Future PRs to improve parsing welcome. :)

PR #236, opened 2020-05-30
(maint) Remove pin to custom version of Litmus

glennsarti:
Previously the litmus gem was pinned to a version in Michael Lombardi's fork.
This commit removes the pin as the PR mentioned was merged into Litmus 0.16.0

Ref - https://github.com/puppetlabs/puppet_litmus/commit/f09c9db745e2fdded9ebc75dec1705e39722acdb


glennsarti:
This is breaking due to https://github.com/puppetlabs/puppet_litmus/issues/300  😢

Issue #242, opened 2020-07-02
text rendering as emojis in strings generated docs


wyardley:
## Describe the Bug
Certain strings (e.g., .`erlang::repo::apt::erlang_solutions`) are rendered without being backtick escaped in certain places (headings, default values), and show up as emojis in rendered docs, as seen in 
thttps://github.com/voxpupuli/puppet-erlang/blob/master/REFERENCE.md

<img width="810" alt="image" src="https://user-images.githubusercontent.com/10714426/86309611-008cae00-bbd1-11ea-9c3f-37a8e980d6b6.png">

## Expected Behavior
The string is rendered without emojis

## Steps to Reproduce
1. Create documentation with a parameter that will produce an emoji (see examples)
2. Use Puppet strings to generate docs
3. ???
4. Profit

## Environment
 - n/a

## Additional Context
https://stackoverflow.com/questions/50657698/how-can-i-stop-github-from-rendering-emojis-in-markdown covers some of the possible workarounds. Wrapping in backticks should work.

binford2k:
fixed in #243

PR #243, opened 2020-07-08
(#242) Wrap names in backticks when rendering to markdown

scotje:
Updates the markdown templates to wrap many resource and value names in
backticks to prevent emoji from being inadvertantly rendered when a
resource name contains a namespace that happens to also be an emoji
sequence on Github et. al.


wyardley:
Nice - thanks!

PR #246, opened 2020-08-06
Handle a missing description gracefully

ekohl:
b5da040babf4bdbbcfd9bfabdb210041ccfa8ce3 introduced word_wrap which assumes text is a string. However, this is not true and not every resource type has a description.


ekohl:
I haven't attached a test since I don't know how this was actually triggered. I found it in https://github.com/theforeman/puppet-pulp/releases/tag/8.1.0 when releasing that particular tag using puppet-strings 2.5.0. This allowed me to proceed.

binford2k:
The test failure is unrelated to this PR

PR #247, opened 2020-08-06
Handle function return tags without a type

ekohl:
Puppet functions can specify a return type. In this case, there's no need to specify a type in the tag. Previously it tag.types was nil and thus calling nil.first.


ekohl:
Still a draft since I haven't had time to actually look into calling the fixture, but that is a reproducer for the behavior.

DavidS:
This has been merged as #268

Issue #248, opened 2020-08-07
Tags (Parameter, Returns) not generated in HTML output


raemer:
## Describe the Bug
Descriptions from @params and @returns are missing in HTML output.

## Expected Behavior
Descriptions from @params and @returns should be visible in HTML output

## Steps to Reproduce
1. Document a (Ruby 4.x API) function and add @param and @returns tags.
2. Run puppet strings generate.

## Environment
 - Version 2.1.0

## Sources of Error and Fix
When adding support for @note and @todo, the yieldall-statement in overview moved from being the last entry, therefor it is not applied to the nesting any more. Moving @note and @todo into the nexting fixes this, I will add a pull request.

scotje:
Resolved by #249

PR #249, opened 2020-08-07
puppet_function template: fix tags, source

raemer:
Move start of nesting to 'note' in setup.rb, otherwise nesting is not affected by yield in overview.


raemer:
Fixes "Tags (Parameter, Returns) not generated in HTML output #248"

scotje:
Thanks for the fix!

PR #266, opened 2020-12-17
Only set tasks = true when parsing plans.

### LINKED ISSUE #251: class with "apply" attribute causes parser error
## Describe the Bug
On a manifest that contains the following code:
```
reboot { 'Patching as Code - Patch Reboot':
  apply    => 'immediately',
  schedule => 'Patching as Code - Patch Window'
}
```
The puppet strings parser errors out with
```
[error]: Failed to parse manifests/init.pp: Syntax error at 'apply' (line: 173, column: 3)
```
If the specific line with `apply =>` gets commented out, the parser succeeds.

## Expected Behavior
The strings parser should not handle occurrences of the term `apply` as a special case, if they are attributes of classes.

## Steps to Reproduce
1. Create a valid puppet manifest
2. Add a resource that contains an attribute named `apply`
3. Run `puppet parser validate` to confirm the manifest is valid
4. Run `puppet strings generate`
5. Notice the parser chokes on the `apply` attribute.

## Environment
 - PE 2019.8.0
 - Client: MacOS, Puppet Agent 6.0.2, puppet-strings 2.4.0, yard 0.9.24

binford2k:
Only set tasks = true when parsing plans. The tests fail because the tested pathname is not the same pattern, so the regex fails. Do you have suggestions @scotje?

Fixes #251


lib/puppet-strings/yard/parsers/puppet/parser.rb
```diff
@@ -22,10 +22,12 @@ def initialize(source, filename)
   # @return [void]
   def parse
     begin
-      Puppet[:tasks] = true if Puppet.settings.include?(:tasks)
-      if Puppet::Util::Package.versioncmp(Puppet.version, "5.0.0") < 0 && @file.to_s.match(/^plans\//)
-        log.warn "Skipping #{@file}: Puppet Plans require Puppet 5 or greater."
-        return
+      if @file.to_s.match(/^plans\//)
```
binford2k, line 25: @scotje this regex fails during testing because it's in the full fixtures directory. If I remove the start-of-string anchor, tests pass, but that's probably not a good solution. Any other ideas?

scotje: We should probably extract this test into a helper method and then we can stub that out in the tests

pmcmaw:
Hey @binford2k just wondering if this is still functionality you are still interested in working on?

binford2k:
@pmcmaw I'll finish it up next week after kickoff

da-ar:
Hello @binford2k, there's probably a little more required now to get the PR across the line. What would you like to do?

lib/puppet-strings/yard/parsers/puppet/parser.rb
da-ar, line 25: Would this work well enough? Original use case is there to only match `plans/` but an additional check for `/plans/` matches for the test without, i believe, hitting any accidental matches for something that isnt a plan?

```suggestion
      if @file.to_s.match(/[^plans|\/plans]\//)
```

PR #281, opened 2021-06-29
Add support for ensurable in types_extras_handler

joshcooper:
Puppet's `file` type uses the `ensurable` method in an extra file, so add
support for that.

### DIFF
--- a/lib/puppet-strings/yard/handlers/ruby/type_extras_handler.rb
+++ b/lib/puppet-strings/yard/handlers/ruby/type_extras_handler.rb
@@ -5,7 +5,7 @@ require 'puppet-strings/yard/handlers/ruby/type_base'
 require 'puppet-strings/yard/code_objects'
 require 'puppet-strings/yard/util'
 
-# Implements the handler for Puppet resource type newparam/newproperty calls written in Ruby.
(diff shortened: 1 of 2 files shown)


joshcooper:
Manually tested that `puppet strings` generates ensure docs for the file type:

```
$ grep ensure doc/puppet_types/file.html 
type should be specified in the <code>ensure</code> attribute.</p>
      <span class="name">ensure</span>
<li><code>absent</code> ensures the file doesn't exist, and deletes it if necessary.</li>
<li><code>file</code> ensures it's a normal file, and enables use of the <code>content</code> or
<li><code>directory</code> ensures it's a directory, and enables use of the <code>source</code>,
<li><code>link</code> ensures the file is a symlink, and <strong>requires</strong> that you also
```

jpogran:
Thanks for the work @joshcooper! Is there a specific issue/JIRA ticket that inspired this (so I can add it to the changelog)? If not I'll merge as is and put up a PR to amend the CHANGELOG with an Unreleased section for future releases.

joshcooper:
Ah sorry forgot to update this. The issue came about in https://tickets.puppetlabs.com/browse/DOCUMENT-1232

PR #295, opened 2022-03-23
add snyk workflow

LivingInSyn:
This PR adds snyk scanning via GitHub action and writes the the Puppet General org

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -24,3 +24,4 @@ jobs:
           SNYK_TOKEN: ${{ secrets.SNYK_GENERAL_KEY }}
         with:
           command: monitor
+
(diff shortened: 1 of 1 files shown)


.github/workflows/snyk.yml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          ruby-version: 2.7
+      - name: create lock
+        run: bundle lock
+      - name: Run Snyk to check for vulnerabilities
+        uses: snyk/actions/ruby@master
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_GENERAL_KEY }}
+        with:
+          command: monitor
```
bastelfreak, line 26: I'm a huge fan of files with a proper \n at the end of the file. This will reduce the diff if another line is added in the future.

LivingInSyn: Good callout, resolved 👍

chelnak (approved):

PR #299, opened 2022-06-09
add data type list when having only data type aliases (fixes #298)

### LINKED ISSUE #298: Docs with Data Type Aliases without Data Type will not generate lists
Documentation containing data type aliases without any data types will not generate puppet_data_type_list.html and the left navigation iframe will contain 404 not found error when on Data Type Alias documentation page. Also the top navigation will be missing Data Types section.

mmolnar:
Documentation containing data type aliases without any data types will not generate puppet_data_type_list.html and the right navigation iframe will contain 404 not found error when on Data Type Alias documentation page. Also the top navigation will be missing Data Types section.

This is caused by only puppet_data_type existence being checked when determining wether to generate the list, but the whole section is setup to contain both puppet_data_type and puppet_data_type_alias.

### DIFF
--- a/lib/puppet-strings/yard/templates/default/layout/html/setup.rb
+++ b/lib/puppet-strings/yard/templates/default/layout/html/setup.rb
@@ -81,11 +81,6 @@ def create_menu_lists
       search_title: 'Puppet Classes'
     },
     {
-      type: 'puppet_data_type',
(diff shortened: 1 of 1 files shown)


mmolnar:
The link to sign the CLA seems to be broken...

Issue #304, opened 2022-09-23
Some values in generated Markdown get double backticks


danielparks:
## Describe the Bug

Some values in generated Markdown are wrapped with double backticks instead of single.

As far as I can tell this doesn’t affect the HTML rendering on GitHub or Forge.

## Expected Behavior

All values should be wrapped with single backticks.

## Steps to Reproduce

```
❯ cat test.pp     
class test ( $param = true ) {}
❯ puppet strings generate --format markdown test.pp &>/dev/null ; grep '^Default value' REFERENCE.md
Default value: ``true``
```

## Environment
 - ruby 3.1.2p20
 - macOS 12.6.0
 - Current main branch (dba16eddabd4947f1901271648097b2f93aa13b5)

---

I’ll submit a PR for this shortly.

PR #305, opened 2022-09-23
(#304) Fix double backticks in Markdown

danielparks:
Previously, some values in generated Markdown were wrapped in double backticks instead of single backticks. This removes the code that added the extra backticks.

See #304.

### DIFF
--- a/lib/puppet-strings/markdown/base.rb
+++ b/lib/puppet-strings/markdown/base.rb
@@ -158,19 +158,6 @@ module PuppetStrings::Markdown
       clean_link(name)
     end
 
-    # Some return, default, or valid values need to be in backticks. Instead of fu in the handler or code_object, this just does the change on the front.
-    # @param value
-    #  any string
-    # @return [String] value or value in backticks if it is in a list
-    def value_string(value)
-      to_symbol = %w[undef true false]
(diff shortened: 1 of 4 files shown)


binford2k:
are there other parts of the codebase that might be affected by this?

danielparks:
I don’t think so. It appears that [the `value_string` function](https://github.com/puppetlabs/puppet-strings/pull/305/files#diff-d839a9e39b18a1d882ec8ee5ca04eef5059d855e29cedde26073615aef6a3090L157-L168) was added (1374b67da0cc89df0949aa84da289d9e10a18c91) before backticks were used for all values, and then it was left in place (07ef00433ce63ce60ffa38438cb229432c5a939a). I removed all calls to it.

It’s possible that some of the litmus acceptance tests will start failing because of this, but I didn’t find any tests for double backticks, or that called `value_string` directly. Unfortunately, it’s pretty easy to accidentally generate Ruby code that doesn’t show up in a search for such things.

Also unfortunately, the litmus tests seem to be broken both here and on my machine. :( I got Litmus working in my own module with vagrant, but it took a lot of digging, so I’m not terribly optimistic that will be easy to figure out without an expert.

chelnak (approved):

Issue #302, opened 2022-09-23
Lots of warnings about when running unit tests


danielparks:
## Describe the Bug

Running `bundle exec rake spec` generates lots of warnings even though the tests run successfully.

Most of the warnings are about two calls to `ERB.new`. The unique warnings are:

> …/puppet-strings/lib/puppet-strings/markdown/table_of_contents.rb:21: warning: Passing safe_level with the 2nd argument of ERB.new is deprecated. Do not use it, and specify other arguments as keyword arguments.
> …/puppet-strings/lib/puppet-strings/markdown/table_of_contents.rb:21: warning: Passing trim_mode with the 3rd argument of ERB.new is deprecated. Use keyword argument like ERB.new(str, trim_mode: ...) instead.
> …/puppet-strings/lib/puppet-strings/markdown/base.rb:186: warning: Passing safe_level with the 2nd argument of ERB.new is deprecated. Do not use it, and specify other arguments as keyword arguments.
> …/puppet-strings/lib/puppet-strings/markdown/base.rb:186: warning: Passing trim_mode with the 3rd argument of ERB.new is deprecated. Use keyword argument like ERB.new(str, trim_mode: ...) instead.

There are also a bunch of “errors” that don’t result in test failures. They are all identical: (**Update:** I split this out into #307)

> \[error\]: Failed to parse (stdin): Syntax error at 'param1' (line: 5, column: 19)

I have tested this with Ruby 2.6.0p0 (not 2.3.0, see #301) and 3.1.2p20. In 2.6.0 the ERB warnings are not present, but the “ Failed to parse” messages are present in both.

## Expected Behavior

No extra output when running unit tests.

## Steps to Reproduce
1. `bundle exec rake spec`

## Environment
 - ruby 3.1.2p20
 - Current main branch (dba16eddabd4947f1901271648097b2f93aa13b5)

danielparks:
After a bunch of testing, it looks like the parse error is at https://github.com/puppetlabs/puppet-strings/blob/dba16eddabd4947f1901271648097b2f93aa13b5/spec/unit/puppet-strings/json_spec.rb#L33

The ERB error, on the other hand, is easy to fix as long as the minimum ruby version is 2.6.0.

danielparks:
I fixed the ERB error for all supported Ruby versions: https://github.com/puppetlabs/puppet-strings/compare/main...danielparks:puppet-strings:fix_erb

chelnak:
[...]

Issue #301, opened 2022-09-23
Minimum Ruby version is incompatible with current code


danielparks:
## Describe the Bug

[puppet-strings.gemspec](https://github.com/puppetlabs/puppet-strings/blob/dba16eddabd4947f1901271648097b2f93aa13b5/puppet-strings.gemspec#L16) reports that the minimum version of Ruby supported is 2.3.0, but there is unsupported code ([`match?`](https://apidock.com/ruby/String/match%3F)) since at least 08936f963e6e0b98c41d66be166b8635f513d18c.

There are also a bunch of unsupported dependency problems, but that can be solved with the brute force addition of a lot of version constraints.

Switching to ruby 2.6.0p0 makes the code work fine without problems.

## Expected Behavior

You should be able to install dependencies and run unit tests with the minimum ruby version.

## Steps to Reproduce
1. `rbenv install 2.3.0 && rbenv local 2.3.0`
2. `gem install bundler:1.13.7` (later versions may work)
3. `bundle install --path .bundle/gems` (this will fail)
4. Add a bunch of version constraints to Gemfile.
5. `bundle install --path .bundle/gems`
6. `bundle exec rake spec`

```
. . .
      RuntimeError:
        Processing (stdin):13 with . . . => undefined method `match?' for /ruby4x/:Regexp
. . .
```

## Environment
 - ruby 2.3.0p0
 - macOS 12.6.0 
 - Current main branch (dba16eddabd4947f1901271648097b2f93aa13b5)

danielparks:
I updated this to better reflect reality. I’m not sure what the implications of updating the minimum version would be.

I can prep a PR to fix all the errors with Ruby 2.3.0, but it involves specifying a bunch of additional dependencies with version information. It will eventually break as dependencies stop supporting 2.3.0, which will require additional constraints to be added, but that’s unavoidable with any Ruby version (perhaps unless Gemfile.lock gets added, though I don’t think that’s supported by bundler with Ruby 2.3.0).

danielparks:
To give you an idea of the number of dependencies that need to be modified: https://github.com/puppetlabs/puppet-strings/compare/main...danielparks:puppet-strings:fix_ruby_2.3.0

I only made very minimal effort to specify them in the correct section; it’s not ready to be a PR.

chelnak:
@danielparks I think it's reasonable to bump the minimum version. I'll discuss internally too just incase anyone can see an issue with it.

danielparks:
[...]

PR #308, opened 2022-09-26
(#302) Fix warnings generated by ERB.new

danielparks:
This has to special case Ruby <2.6.0 since those versions will interpret the keyword argument as a hash passed as the second argument.

Without this change running this code in Ruby 2.6+ will generate warnings about the parameters passed to `ERB.new`.

---

I’ve run the tests under Ruby 2.3.0 (with some dificulty; see #301) and this appears to work. At the very least it clears up all the warnings.

### DIFF
--- a/lib/puppet-strings/markdown/base.rb
+++ b/lib/puppet-strings/markdown/base.rb
@@ -172,9 +172,9 @@ module PuppetStrings::Markdown
 
     # @return [String] full markdown rendering of a component
     def render(template)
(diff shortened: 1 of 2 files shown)


bastelfreak (approved):

bastelfreak:
@chelnak since this causes a lot of noise on ruby3, could you do a new release please?

chelnak:
Ideally I would like to cut a new release for this today.

There are some other clean-up tasks that I want to sort out first though.

Issue #307, opened 2022-09-25
Parsing plans in unit tests fail without being caught


danielparks:
## Describe the Bug

It currently fails to parse the plan in [spec/unit/puppet-strings/json_spec.rb](https://github.com/puppetlabs/puppet-strings/blob/dba16eddabd4947f1901271648097b2f93aa13b5/spec/unit/puppet-strings/json_spec.rb#L28-L35) and fails with error `[error]: Failed to parse (stdin): Syntax error at 'param1' (line: 5, column: 19)`.

Adding a new unit test to parse a plan also fails.

However, it is able to parse plans in the wild, e.g. https://github.com/puppetlabs/puppetlabs-pam_tools/blob/main/REFERENCE.md#plans-1 . I suppose it’s possible that it has stopped working since that was generated; I haven’t yet checked.

This is an offshoot of #302; it seemed like this might be a bigger issue than just warnings during tests so I figured I should split it out.

## Expected Behavior

It should parse a plan successfully without generating failures and generate documentation for it.

## Steps to Reproduce
1. `bundle exec rake spec`

## Environment
 - ruby 3.1.2p20
 - macOS 12.6.0
 - Current main branch (dba16eddabd4947f1901271648097b2f93aa13b5)

danielparks:
I was sort of wrong about the second test. It was only failing because I was only running one file. Running all the tests made it work.

It turns out I can fix the original error by running this code at the top of the file (spec/unit/puppet-strings/json_spec.rb):

```ruby
YARD::Parser::SourceParser.parse(File.expand_path("../../fixtures/plans/plan.pp", __dir__))
PuppetStrings::Markdown.generate
```

WTF.

danielparks:
Oh, or alternatively,

```ruby
YARD::Parser::SourceParser.parse(File.expand_path("../../fixtures/plans/plan.pp", __dir__))
YARD::Registry.clear
```

To be clear, both lines are required to make it work, though they don’t have to be in that order.

danielparks:
Looks like this issue has something to do with the parser checking file paths to see if they match `%r{^plans|/plans/}`. If they do it might set `Puppet[:tasks] = true`. https://github.com/puppetlabs/puppet-strings/blob/dba16eddabd4947f1901271648097b2f93aa13b5/lib/puppet-strings/yard/parsers/puppet/parser.rb#L25-L31

[...]

PR #311, opened 2022-09-26
(#307) Fix plan parsing in JSON spec tests

danielparks:
This is 3 closely related commits. I’m not totally satisfied with this solution — I would be inclined to set `Puppet[:tasks] = true` regardless of the file name, but I’m not really sure what all the implications would be.

### Catch parse failures in `PuppetStrings::Json` tests

Previously, the tests for `PuppetStrings::Json` (in json_spec.rb) parsed a variety of code samples and then queried YARD to test the results. If a sample failed to parse, an error might be outputted (see [issue #307][#307]), but no failure was registered by the testing harness.

This adds simple checks to verify that the parses actually succeed.

Unfortunately, one of the samples currently fails to parse, so this commit introduces a testing error. The subbsequent commit will fix it.

[#307]: https://github.com/puppetlabs/puppet-strings/issues/307

### (#307) Fix plan parsing in JSON spec tests

Previously, parsing of Puppet plan code failed in json_spec.rb because it didn’t provide a file name to the parser, which prevented the parser from enabling task support in Puppet, which prevented the parser from recognizing a plan.

This turns on task support in Puppet before trying to parse a plan in json_spec.rb.

### Add a plan test to parser_spec.rb

Previously, none of the examples in parser_spec.rb included a Puppet plan. This adds a plan and verifies that it parses correctly.

### DIFF
--- a/spec/unit/puppet-strings/json_spec.rb
+++ b/spec/unit/puppet-strings/json_spec.rb
@@ -25,6 +25,8 @@ define dt(Integer $param1, $param2, String $param3 = hi) {
 }
     SOURCE
 
+    # Puppet will not parse plans if Puppet[:tasks] is not enabled.
+    Puppet[:tasks] = true if TEST_PUPPET_PLANS
     expect(YARD::Parser::SourceParser.parse_string(<<-SOURCE, :puppet).enumerator.length).to eq(1) if TEST_PUPPET_PLANS
 # A simple plan.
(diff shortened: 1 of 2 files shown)


danielparks:
I’m inclined to close this in favor of #312, but I figured I’d let somebody else make the call. Sorry for the noise!

danielparks:
Closing in favor of #312.

PR #313, opened 2022-09-26
(#301) Update minimum Ruby version to 2.5.0

danielparks:
Previously, the minimum Ruby version was 2.3.0, but it was dificult to get it working and some of the code had been updated to use features as new as 2.4.6.

This clears out code that checked for old Ruby versions, and removes some version constraints in Gemfile that were needed to work with Ruby 2.3.0.


chelnak (approved):
❤️

Issue #267, opened 2021-01-05
`Public Functions` header shouldn't be output if there are only private functions


alexjfisher:
## Describe the Bug
`Public Functions` header shouldn't be output if there are only private functions

## Expected Behavior
When a function is marked as `@api private` it appears under a `Private Functions` header instead of the generic `Functions` header.  If there are no public functions, the `Public Functions` header should be omitted.

See example and conversation [here](https://github.com/voxpupuli/puppet-chrony/pull/103/files#r552030356).

danielparks:
This is also true for private resource types, and possibly other things as well.

danielparks:
PR #314 submitted.

danielparks:
Should be fixed now in main!

PR #310, opened 2022-09-26
(MAINT) Add CI workflow

chelnak:
Prior to this commit puppet-strings relied on travis for ci. We have since moved away from travis and use GitHub actions.

This commit creates a new ci.yml workflow that runs on:

* A nightly cron
* pull request to main
* workflow_dispatch

We also remove the legacy travis workflow.

In addition to the above there are also a number of rubocop fixes in this PR. These fixes were required to ensure that spec and linting tests were green.

PR #314, opened 2022-09-26
(#267) Don’t show “Public X” header without contents

danielparks:
Previously, the “Public X” header would be displayed if there were any private X regardless of whether there were any public X.

For example, having a private function would make both the “Public Functions” and “Private Functions” headers appear, even if there weren’t any public functions.

This changes it so that there are three conditions:

  * **Only public X:** no “Public” or “Private“ headers are displayed; the X are listed with links to their documentation. (Public is implied.)
  * **Only private X:** a “Private X” header is is displayed with a list of X. There are no links to documentation for private APIs.
  * **Both public and private X:** both “Public X” and “Private X” headers are displayed. The public Xs are listed with links to their documentation; the private Xs are just listed with no links.

In other words, if there are no private Xs, then it just lists the public once. Otherwise, it splits them up under public/private headers but avoids showing a header if it’s contents will be empty.

This also radically simplifies and removes a bunch of boilerplate code around rendering these sections.


lib/puppet-strings/markdown/table_of_contents.rb
```diff
@@ -3,24 +3,17 @@
 module PuppetStrings::Markdown
   module TableOfContents
     def self.render
-      final = "## Table of Contents\n\n"
+      template = File.join(File.dirname(__FILE__), 'templates/table_of_contents.erb')
```
ekohl, line 6:
```suggestion
      template = File.join(__dir__, 'templates', 'table_of_contents.erb')
```

lib/puppet-strings/markdown/table_of_contents.rb
```diff
@@ -3,24 +3,17 @@
...
-      PuppetStrings::Markdown::PuppetTasks,
-      PuppetStrings::Markdown::PuppetPlans].each do |r|
-        toc = r.toc_info
-        group_name = toc.shift
-        group = toc
-        priv = r.contains_private?
-
-        template = File.join(File.dirname(__FILE__), 'templates/table_of_contents.erb')
-        final += PuppetStrings::Markdown.erb(template).result(binding)
+        PuppetStrings::Markdown.erb(template).result(binding)
```
ekohl, line 13: I wonder if you can also store this outside of the loop (just like the `template` variable) and render it multiple times

ekohl (commented):
Nice deduplication. Looks like there's room for more. Some minor details inline.

lib/puppet-strings/markdown/table_of_contents.rb
danielparks: Good question!

ekohl (approved):

danielparks:
I realized that it was a lot clearer to pull `PuppetStrings::Markdown::TableOfContents` into the general renderer, so I did. I also did some future-rubocop-related clean ups, though it’s still going to take some work to rebase this after #310 lands.

spec/unit/puppet-strings/markdown_spec.rb
```diff
@@ -2,56 +2,55 @@
...
-    YARD::Parser::SourceParser.parse_string(fixture_content("ruby/resource_type.rb"), :ruby)
-    YARD::Parser::SourceParser.parse_string(fixture_content("ruby/resource_api.rb"), :ruby)
-
-    # task metadata derives the task name from the filename, so we have to parse
-    # directly from the filesystem to correctly pick up the name
-    YARD::Parser::SourceParser.parse(File.join(fixture_path, "json/backup.json"))
-  end
+    def parse_shared_content
+      # Populate the YARD registry with both Puppet and Ruby source
+      YARD::Parser::SourceParser.parse_string(fixture_content("puppet/class.pp"), :puppet)
```
ekohl, line 20: Isn't it tempting to rewrite this to a loop?

lib/puppet-strings/markdown/data_types.rb
```diff
@@ -3,41 +3,19 @@
 require_relative 'data_type'
 
 module PuppetStrings::Markdown
+  # Adapter to get information about data types
   module DataTypes
```
ekohl, line 7: I was really wondering if this class shouldn't be killed off instead. Just merge it into `DataType`. Possibly better for a future PR.

ekohl (approved):
So tempting to remove even more code.

lib/puppet-strings/markdown.rb
```diff
@@ -41,4 +70,17 @@ def self.render(path = nil)
       YARD::Logger.instance.debug "Wrote markdown to #{path}"
     end
   end
+
+  # Helper function to load an ERB template.
+  #
+  # @param [String] path The full path to the template file.
+  # @return [ERB] Template
+  def self.erb(path)
+    if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.6.0')
```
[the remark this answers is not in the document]
danielparks: Hm… I thought about it, but I’m inclined not to since it ends up duplicating the doc comment. Or maybe I should just put the docs above the outer `if` if I do that? Pretty sure the performance difference will be negligible.

spec/unit/puppet-strings/markdown_spec.rb
danielparks: Ha! This has already started snowballing a bit… I keep coming back and finding little tweaks, and then those bring more things in scope that I could tweak…

lib/puppet-strings/markdown/data_types.rb
danielparks: Ah ha! I’d been thinking about different ways to clean this up, but for some reason that didn’t occur to me. It makes it a lot cleaner.

bastelfreak (approved):

chelnak (approved):

lib/puppet-strings/markdown/base.rb
```diff
@@ -50,6 +50,33 @@ module PuppetStrings::Markdown
   #  ") inherits foo::bar {\n" +
   #  "}"}
   class Base
+    # Set or return the name of the group
+    #
+    # @param [Optional[String]] Name of the group to set
+    # @return [String] Name of the group
+    def self.group_name(name = nil)
```
ekohl, line 57: Thoughts about splitting the functionality in `self.group_name()` and `self.group_name=(name)`? Would that work?

Same comment applies to `yard_types`.

lib/puppet-strings/markdown/base.rb
```diff
@@ -50,6 +50,33 @@ module PuppetStrings::Markdown
...
+      @yard_types
+    end
+
+    # @return [Array] list of items
+    def self.items
+      yard_types
+        .flat_map { |type| YARD::Registry.all(type) }
+        .sort_by(&:name)
+        .map(&:to_hash)
+        .map { |i| new(i) }
```
ekohl, line 77: You can combine both maps
```suggestion
        .map { |i| new(i.to_hash) }
```

lib/puppet-strings/markdown/base.rb
```diff
@@ -50,6 +50,33 @@ module PuppetStrings::Markdown
...
+    # @return [Array[Symbol]] Array of symbols registered with YARD
+    def self.yard_types(types = nil)
+      @yard_types = types if types
+      @yard_types
+    end
+
+    # @return [Array] list of items
+    def self.items
+      yard_types
+        .flat_map { |type| YARD::Registry.all(type) }
```
ekohl, line 74: TIL about https://ruby-doc.org/core-2.7.1/Enumerable.html#method-i-flat_map. Neat!

lib/puppet-strings/markdown/base.rb
danielparks: I actually tried doing

```ruby
class << self
  attr_accessor :group_name
  attr_accessor :yard_types
end
```

but I ran into a few problems. One was that it interpreted `yard_types = [...]` as trying to set a local variable instead of calling a class method, but maybe I could have called `self.yard_types = [...]`.

I wanted it to look more like functions like `attr_accessor` since I figured that would feel more natural, but it did seem inconsistent with the way Ruby does things elsewhere.

I don’t have a real good sense for the Ruby Way anymore… been writing other languages for too long.

lib/puppet-strings/markdown/base.rb
danielparks: 🤦

lib/puppet-strings/markdown/base.rb
ekohl: Fair enough. I suppose this is good enough.

Issue #240, opened 2020-06-15
Default value expressions are returned incorrectly


nicklewis:
## Describe the Bug
Complex parameter default expressions (specifically those involving an AST branch node) are returned incorrectly.

For example:
```puppet
class test($param = 1 + 1) { ... }
```

The default value for `$param` is returned as `+`.

## Expected Behavior
The default value should be returned as `1 + 1`.

## Additional Context

This seems to happen because `ParameterizedStatement` parser uses a `SourcePosAdapter` from Puppet to extract the text for the AST object. However, the `extract_text` method it calls only extracts the text for the AST branch itself, not for the entire tree rooted in that branch. The `SourcePosAdapter` calls  `@object.locator.extract_text`, but there is an `extract_tree_text` method available on the locator which seems to do the right thing.

https://github.com/puppetlabs/puppet-strings/blob/main/lib/puppet-strings/yard/parsers/puppet/statement.rb#L89-L90
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pops/adapters.rb#L63
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pops/parser/locator.rb#L55

I would file a PR but I'm not deep enough into puppet-strings to understand its relationship to Puppet and the compatibility issues involved with the `SourcePosAdapter` to know the appropriate place to make this change.

seanmil:
I encountered this with a default expression like `$facts.dig('key', 'subkey')`, the parsed default value is only `('key', 'subkey')`. I assume this the same root issue.

danielparks:
I have also encountered this bug.

danielparks:
Well, `extract_tree_text` _almost_ works. It fails for the default value of `$param1.strip()` — it gets `$param1.strip(` instead. I suspect that’s a bug in Puppet.

You can try my branch: https://github.com/puppetlabs/puppet-strings/compare/main...danielparks:puppet-strings:fix_default_expressions (That includes a fix for the ERB warnings as well, though it will break on Ruby <2.6, I think, so it can’t be committed. See #302)

danielparks:
Filed ticket https://tickets.puppetlabs.com/browse/PUP-11632 (not sure that’s the right project, but ¯\\\_(ツ)\_/¯)

danielparks:
Partially fixed — it’s much better, but sometimes [PUP-11632](https://tickets.puppetlabs.com/browse/PUP-11632) strips the final character.

PR #315, opened 2022-09-27
(#240) Fix output of default values that are expressions

danielparks:
Previously, parameters with a default value that was an expression were not outputted into the documentation correctly. For example,

    Integer $param = 1 + 1,

Would be shown in the documentation to have a default value of “+”.

This switches to using `extract_tree_text` instead of `extract_text` to get the text representation of the parsed Puppet code.

This also gets rid of the dependency on `Puppet::Pops::Adapters::SourcePosAdapter`, which was [deprecated in 2017](https://github.com/puppetlabs/puppet/commit/68498adfa2560c9cbf28d7a46ebc1aef14c69992).

Unfortunately, it appears that there is a bug in Puppet ([PUP-11632][]) that sometimes returns the incorrect default value. This adds two skipped tests that demonstrate the issue.

[PUP-11632]: https://tickets.puppetlabs.com/browse/PUP-11632

### DIFF
--- a/lib/puppet-strings/yard/handlers/ruby/data_type_handler.rb
+++ b/lib/puppet-strings/yard/handlers/ruby/data_type_handler.rb
@@ -152,12 +152,12 @@ class PuppetStrings::Yard::Handlers::Ruby::DataTypeHandler < PuppetStrings::Yard
 
     def literal_AccessExpression(o)
       # Extract the raw text of the Access Expression
-      ::Puppet::Pops::Adapters::SourcePosAdapter.adapt(o).extract_text
+      PuppetStrings::Yard::Util.ast_to_text(o)
     end
 
     def literal_QualifiedReference(o)
       # Extract the raw text of the Qualified Reference
-      ::Puppet::Pops::Adapters::SourcePosAdapter.adapt(o).extract_text
+      PuppetStrings::Yard::Util.ast_to_text(o)
     end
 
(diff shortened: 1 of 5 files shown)


danielparks:
I was going to let this sit until I heard back about [PUP-11632](https://tickets.puppetlabs.com/browse/PUP-11632), but I figured that this was an improvement even with the bug.

chelnak:
I like this change. I'm just trying to weigh up the behaviour change. 

Does it put users in a better position? Or would this only occur some of the time?

danielparks:
You’re asking how often the [PUP-11632](https://tickets.puppetlabs.com/browse/PUP-11632) bug occurs and gives a bad default value?

It seems fairly rare. So far I’ve only found two cases, which are both documented in the bug and in the unit tests attached to this PR.

It seems to have much less impact than the bug this PR fixes. As is, you often get what appears to be an incomprehensible fragment of the default value, e.g. `+` instead of `1 + 1`, or `(': ')[0]` instead of `split($name, ': ')[0]`.

With this PR you get a few munged values, but so far it’s only been a missing final character, e.g. `$foo.func(` instead of `$foo.func()`.

chelnak:
@danielparks thanks for the explanation .. it's much clearer now.

I'm happy to get this merged!

chelnak (approved):

PR #316, opened 2022-09-29
(MAINT) Add release workflows

chelnak:
Prior to this PR releasing `puppet-strings` was a manual effort.

This PR adds two new workflows:

## release_prep.yml

This workflow will update the changelog based on the version specified in the current gemspec.
It does not automatically generate the next version number. This will need to be done in a commit before this workflow is executed.

## release.yml

This workflow will build a new gem, create a GitHub release (and tag), upload the gem as a release asset and finally push the gem to rubygems.org.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,16 @@
+name: "Release"
+
+on:
+  workflow_dispatch:
+   inputs:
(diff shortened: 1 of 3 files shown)


GSPatton (approved):
all looks good

PR #317, opened 2022-09-29
Use tilde heredocs for readability

danielparks:
This changes most heredocs to use the tilde syntax with variable interpolation off. It indents the contents of the heredoc along with the final symbol.

To maintain these changes, it enables `Layout/HeredocIndentation` and `Layout/ClosingHeredocIndentation` in Rubocop.

The tilde syntax smartly strips the indentation, and allows for different indention of the closing identifier:

```ruby
"test\n" == <<~'END'
    test
  END
```

This syntax was introduced in Ruby 2.3.0, so this will not pose compatibility problems.

### DIFF
--- a/spec/unit/puppet-strings/yard/handlers/ruby/function_handler_spec.rb
+++ b/spec/unit/puppet-strings/yard/handlers/ruby/function_handler_spec.rb
@@ -19,12 +19,10 @@ describe PuppetStrings::Yard::Handlers::Ruby::FunctionHandler do
 
   describe 'parsing 3.x API functions' do
     describe 'parsing a function with a missing docstring' do
-      let(:source) do
-        <<-SOURCE
-Puppet::Parser::Functions.newfunction(:foo) do |*args|
-end
+      let(:source) { <<~'SOURCE' }
+        Puppet::Parser::Functions.newfunction(:foo) do |*args|
+        end
       SOURCE
-      end
 
       it 'logs a warning' do
         expect { spec_subject }.to output(%r{\[warn\]: Missing documentation for Puppet function 'foo' at \(stdin\):1\.}).to_stdout_from_any_process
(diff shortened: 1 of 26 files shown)


spec/unit/puppet-strings/yard/handlers/ruby/function_handler_spec.rb
```diff
@@ -19,33 +19,30 @@
...
-Puppet::Parser::Functions.newfunction(:foo, doc: <<-DOC
-An example 3.x function.
-@param [String] first The first parameter.
-@param second The second parameter.
-@return [Undef] Returns nothing.
-DOC
-) do |*args|
-end
+      let(:source) { <<~'SOURCE' }
+        Puppet::Parser::Functions.newfunction(:foo, doc: <<~'DOC') do |*args|
```
danielparks, line 34: Looks like I found a weird bug. Putting `) do |*args|` on the same line as the heredoc rather than after the final symbol (as it was previously) causes some number of characters to be trimmed from the docstring before YARD parses the tags (I think).

If you replace `|*args|` with `|*|` then 4 fewer characters will be trimmed.

This shows up as the `@return` tag having trimmed text, but if you put one of the `@param`s on the end, then that will have trimmed text instead. (Which is why I think it happens before it tries to look for tags.)

I’m guessing this is bug in YARD or… the Ruby parser? I’m not sure what YARD uses under the hood. I’ll try to figure out where this should be filed tomorrow.

danielparks:
Updated the tests to work and added lengthy comments (maybe I shouldn’t wrap at 80 characters?).

I also updated the commit message in hope that it will help some future fool:

> ### Tests that should fail
> 
> Note that two tests couldn’t be updated to a more common heredoc format to avoid failing.
> 
> It appears there is a problem with the way YARD parses parameters with defaults with a heredoc ([YARD #779][]). There is a work-around for that issue in `PuppetStrings::Yard::Handlers::Ruby::Base`, but it can’t work in these cases because YARD doesn’t provide enough source to read the whole heredoc content.
> 
> Given that these are tests of the Puppet 3 style functions, it’s probably not worth pursuing this issue.
> 
> [YARD #779]: https://github.com/lsegal/yard/issues/779

As I said, I’m inclined to ignore this, even though it pains me deep inside.

spec/unit/puppet-strings/yard/handlers/ruby/function_handler_spec.rb
```diff
@@ -19,31 +19,37 @@
...
-end
+      # Bug: Putting `) do |*args|` on the first line rather than right after
+      # the end of the heredoc block causes the docstring to be trimmed. This is
+      # probably related to https://github.com/lsegal/yard/issues/779. The code
+      # in PuppetStrings::Yard::Handlers::Ruby::Base has a special work around
+      # for this, but it seems to be limited by how much data YARD provides, and
+      # it often isn’t enough.
+      #
+      # Given that this occurs only in old-style functions, it’s probably not
+      # worth pursuing.
```
danielparks, line 41: One of the problematic tests.

chelnak:
@danielparks Might be worth rebasing this PR when you get five minutes. Then we can merge it.

chelnak:
💯  @danielparks  💯

Issue #223, opened 2020-02-26
Wrong enclosure for data type with multiple lines at markdown


crazymind1337:
## Describe the Bug
If I add a data type with mutliple line, it will be enclosured with single "\`" (inline code section) instead of tripple "\`" (multi-line code fences). This causes a strange view at a markdownv viewer.

## Expected Behavior
It should be enclosed with ```\n<text>\n``` 

Expected result:
```markdown
Alias of

\```
Hash[Pattern[/^[a-z][a-z0-9_-]*$/], Struct[
    {
      param1 => String[1],
      param2 => Stdlib::Absolutepath,
      paramX => Boolean,
    }
  ]
]
\```
```

## Steps to Reproduce
1. add a custom data type, e.g.:
```puppet
# summary Test Type
#
type Test_module::Test_type = Hash[
  Pattern[/^[a-z][a-z0-9_-]*$/],
  Struct[
    {
      param1 => String[1],
      param2 => Stdlib::Absolutepath,
      paramX => Boolean,
    }
  ]
]
```
2. generate the REFERENCE.md:
```shell
puppet strings generate --format markdown --out REFERENCE.md
```
3. check the result, it will be like:
```markdown
Alias of `Hash[Pattern[/^[a-z][a-z0-9_-]*$/], Struct[
    {
      param1 => String[1],
      param2 => Stdlib::Absolutepath,
      paramX => Boolean,
    }
  ]]`
```

## Environment
 - Gem puppet-strings (2.3.1)
 - Platform MacOS 10.15.3 (19D76)

binford2k:
To be clear, this generates markdown with inline code markup instead of code fences.

binford2k:
I've reproduced this issue with the example case provided. Thanks for letting us know @crazymind1337!

danielparks:
Looks like this is already in place for data type aliases, but it would be nice to extend that to default values and data types where we can. I have a branch that fixes this: https://github.com/danielparks/puppet-strings/compare/better_public_private...danielparks:puppet-strings:multiline_code_blocks (it’s waiting on PR #314, my unsubmitted fix for #240, and possibly [PUP-11632](https://tickets.puppetlabs.com/browse/PUP-11632)).

Also, thanks @crazymind1337, I was surprised that I couldn’t just use a multiline `Variant` type as a test case, so I used your example.

danielparks:
[...]

PR #319, opened 2022-09-29
(#223) Use code blocks as appropriate in Markdown

danielparks:
Sometimes data types, default values, and other code outputted in Markdown documentation is multiline. When that’s the case, it’s cleaner to use a code block (below) rather and inline code (`code`).

```
example code block
with two lines
```

This updates the Markdown template code to use code blocks in many places if the code in question contains a newline.

Thanks to @crazymind1337 for the example of a multiline type alias!


chelnak:
@danielparks You are on fire 🔥 

I'll review this as soon as I can

danielparks:
No hurry!

chelnak (approved):
💯

PR #320, opened 2022-09-29
(MAINT) Fix snyk workflow

chelnak:
Prior to this commit an incorrectly configured twingate step was causing the workflow to fail.

We don't actually need twingate so this commit removes the step and hopefully allows it to continue scanning.


GSPatton (approved):
looks good

PR #324, opened 2022-10-10
(MAINT) Add matrix for acceptance tests

chelnak:
This PR preps puppet-strings to use the new reusable GitHub actions. We've surface the matrix in calling workflows now and also changed the way the acceptance tests work. When calling the reusable acceptance workflow, it expects that the caller implements a rake task called acceptance with the required steps inside.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: "ci"
 on:
   push:
     branches:
-      - "maint"
+      - "main"
   pull_request:
     branches:
       - "main"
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -12,6 +12,19 @@ on:
   workflow_dispatch:
 
 jobs:
+
   ci:
-    uses: "puppetlabs/cat-github-actions/.github/workflows/gem_ci.yml@main"
+    uses: "puppetlabs/cat-github-actions/.github/workflows/gem_ci.yml@maint-generic_gem_task"
```
GSPatton, line 17: what is the reason for referencing a specific branch? Is this one of the benefits of cat-github-actions? That we can use different versions of the same workflow by pointing to different branches?

Rakefile
```diff
@@ -177,3 +146,30 @@ rescue LoadError
     raise 'Install github_changelog_generator to get access to automatic changelog generation'
   end
 end
+
+desc 'Run acceptance tests'
+task :acceptance do
+
+  begin
+    if ENV['MATRIX_TARGET']
```
GSPatton, line 154: out of interest, will `MATRIX_TARGET` be one of `puppet6/7` from ci.yml?

.github/workflows/ci.yml
chelnak: It is indeed.. however this PR is not ready for review yet - it should be draft but I need the jobs to trigger (easily).

Rakefile
chelnak: Correct

GSPatton (approved):
looks good to me

PR #327, opened 2022-10-13
(MAINT) Bump gem version

chelnak:
This commit bumps the puppet-string gem version to v3.0.0.


david22swan:
Shouldn't there be a changelog update with this?

david22swan (approved):
LGTM

PR #328, opened 2022-10-21
(MAINT) Pin MDL to 0.11.0

chelnak:
It looks like there was a version bump of mdl this week which introduced some markdown violations.

This commit pins mdl to 0.11.0 to give us time to investigate the changes and keep ci working.


GSPatton (approved):
looks good

PR #342, opened 2023-03-10
Add deprecated tag

### LINKED ISSUE #339: Please add support for deprecated tag
## Use Case
I [recently tried](https://github.com/voxpupuli/puppet-dnsquery/pull/126#pullrequestreview-1303074204) to use the [deprecated](https://www.rubydoc.info/gems/yard/file/docs/Tags.md#deprecated) yard tag hoping that this would create some type of deprecated label on the resulting documentation however as far as i can tell it had no affect.  Could we please add support for the deprecated tag.

## Describe the Solution You Would Like
When using the `@deprecated` tag mark functions in the resulting documentation with a deprecated header e.g.

```ruby
# some function
Puppet::Functions.create_function(:deprecated_function) do
  # @deprecated this function causes issues
 ...
end
```
should result in something like 

```
### <a name="deprecated_function"></a>`deprecated_function`

Type: Ruby 4.x API

[DEPRECATD]  this function causes issues

some function
...
```

b4ldr:
I had a stab at adding the deprecated flag

Fixes #339

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,11 +20,12 @@ group :development do
   gem 'json_spec', '~> 1.1', '>= 1.1.5'
 
   gem 'mdl', '~> 0.11.0'
-  gem 'mocha'
 
   gem 'pry', require: false
   gem 'pry-byebug', require: false
   gem 'pry-stack_explorer', require: false
+  # Need the following otherwise we end up with puppetlabs_spec_helper 1.1.1
+  gem 'mocha', '~> 1.0'
   gem 'puppetlabs_spec_helper'
 
   gem 'rake'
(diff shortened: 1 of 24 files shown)


chelnak:
This looks good and tests are passing.

Just one thing I noticed that we can maybe tidy up.

The description being used for deprecated is "Use a something else".

Could you possibly change the message a bit so it's more relevant to the deprecation of something and a little softer?

Maybe along the lines of "no longer supported and will be removed in a future release".

Thanks!

b4ldr:
thanks for the comments

> Maybe along the lines of "no longer supported and will be removed in a future release".
done

chelnak:
Looks like you have a couple of rubocop violations to clear up.

Once they are done I'm happy to get this merged!

b4ldr:
updated hopefully all green now

chelnak (approved):

chelnak:
@b4ldr Thank you!

PR #360, opened 2023-07-27
(maint) - fix rubocop

jordanbreen28:
https://msp-greg.github.io/rubocop/RuboCop/Cop/Style/ArgumentsForwarding.html


GSPatton (approved):

donoghuc:
@jordanbreen28 @GSPatton I do not believe this is strictly a ruby 2.7 compatible feature. This gem claims support for ruby 2.7 and delegation (`...`) is not supported with leading arguments in for example ruby 2.7.0 https://scriptday.com/blog/2020/09/12/ruby-3-0-added-leading-arguments-forwarding I think we need to revert this in order to maintain ruby 2.7 compatibility (i would expect breaking ruby compatability to be a major version bump in semver).

PR #361, opened 2023-08-04
Fix option tag handling with no data types

seanmil:
Do not error if no data types are provided to the option tag. Instead, generate a suitable line excluding the data type.

## Summary

When using the `@option` tag if a data type is not specified then rendering to Markdown will result in an exception being raised.  This fixes that so an appropriate documentation line is emitted instead.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [x] Manually verified.


jordanbreen28:
Thanks for this @seanmil!

jordanbreen28 (approved):

PR #378, opened 2024-01-04
Drop outdate Ruby version checks

ekohl:
See individual commits for details, but they both come down to checking `RUBY_VERSION` for some old version while the gemspec declares 2.7 as the lower bound.


jordanbreen28 (approved):
thanks @ekohl!

PR #382, opened 2024-01-19
Allow numerics for templates using code_maybe_block

seanmil:
A number of ERB templates use the code_maybe_block method to assist with formatting. This fails if the value being passed is not a string, such as an integer default value for a resource type.

## Summary
Convert the value passed to code_maybe_block into a string before testing for `\n` with include?

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [x] Manually verified.


jordanbreen28:
hey @seanmil this is great. Nice spot and simple fix! Could you rebase of main? Apologies, just merged another PR.

jordanbreen28 (approved):

PR #388, opened 2024-02-27
validate: give hint on how to regenerate outdated REFERENCE.md

kenyon:
This is the command we use with @voxpupuli modules ([ref](https://voxpupuli.org/docs/how_to_run_tests/#referencemd-update)), not sure what the correct way is for PDK modules.

### DIFF
--- a/lib/puppet-strings/tasks/validate.rb
+++ b/lib/puppet-strings/tasks/validate.rb
@@ -34,7 +34,7 @@ namespace :strings do
       existing = File.read(filename)
 
       if generated != existing
-        warn "#{filename} is outdated"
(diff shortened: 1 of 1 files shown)


lib/puppet-strings/tasks/validate.rb
```diff
@@ -34,7 +34,7 @@
       existing = File.read(filename)
 
       if generated != existing
-        warn "#{filename} is outdated"
+        warn "#{filename} is outdated; to regenerate: bundle exec rake strings:generate:reference"
```
bastelfreak, line 37: I am wondering how many people might call puppet strings not via bundler 🤔 but I am +1 for the change

bastelfreak (approved):

PR #397, opened 2024-06-13
Revert "(maint) Exclude breaking rubocop versions"

gavindidrichsen:
Reverts puppetlabs/puppet-strings#393


gavindidrichsen:
As I didn't add any functionality here but only correct syntax on existing code, I'm recommending we

david22swan (approved):

PR #401, opened 2024-09-05
(bug) - Pin yard to < 0.9.37

jordanbreen28:
## Summary
Yard 0.9.37 was released, which broke our pipeline and also broke some module CI.

This commit adds a temporary pin to the gem to < 0.9.37, until we can resolve these issues.

## Additional Context
Add any additional context about the problem here. 
- [ ] Root cause and the steps to reproduce. (If applicable)
- [ ] Thought process behind the implementation.

## Related Issues (if any)
Mention any related issues or pull requests.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.

PR #403, opened 2024-10-31
Fix CI with latest yard

smortex:
Monkey patching libraries is a bad idea.  Latest version of yard changed
some internal details, which broke CI.

A workaround was proposed in #401 but the root cause was not addressed.

In this commit, we replace the private `io` object with an instance of a
class that provide the expected `IO#write` method instead of `nil` which
does not provide this interface in CI, so that the issue is not raised
anymore.

### DIFF
--- a/puppet-strings.gemspec
+++ b/puppet-strings.gemspec
@@ -23,6 +23,6 @@ Gem::Specification.new do |s|
   s.files = Dir['CHANGELOG.md', 'README.md', 'LICENSE', 'lib/**/*', 'exe/**/*']
 
   s.add_runtime_dependency 'rgen', '~> 0.9'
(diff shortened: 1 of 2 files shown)


smortex:
Cc @jordanbreen28 as he authored the workaround and will probably be interested in a fix.

bastelfreak:
@smortex nice workaround! can you take a look at the rubocop complaints?

smortex:
> @smortex nice workaround! can you take a look at the rubocop complaints?

Dang!  Amended, thanks!  Needs approval again to run CI :grin:

jordanbreen28 (approved):

jordanbreen28:
thanks @smortex!!

PR #408, opened 2025-05-30
Rake task allows for a different commit message

ghoneycutt:
Without this commit, you cannot inform the rake task to push gh-pages to accept a different commit message. This is necessary in environment where commit messages have to be in specific formats.

None of the system calls were checking the exit status of their respective commands. Not stopping on error means that the rake take can fail and you and your pipelines would not know.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,8 +15,17 @@ AllCops:
   TargetRubyVersion: '3.1'
 
 # Disabled
+Layout/LineLength:
+  Max: 200
+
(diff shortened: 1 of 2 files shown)


ghoneycutt:
I tested this works by configuring a rake task to use a different message as well as calling the normal rake tasks here and getting the default message as shown in https://github.com/ghoneycutt/puppet-module-pam/pull/279

bastelfreak (approved):

jstraw (approved):

PR #414, opened 2025-09-16
Remove strict puppet dependency

binford2k:
This should no longer depend on `puppet` since there are multiple implementations that can use this. It should not materially affect any workflows since nobody uses `gem install puppet-strings` to install Puppet!


LukasAud (approved):
LGTM

